Issue #4, opened 2015-10-21
Testing Charts on commit


technosophos:
Stories:

As a contributor, I'd like to have my chart automatically tested when I issue a pull request.

As a Helm Steering Committee member, I'd like to know that any charts I am reviewing have passed CI

sgoings:
Currently have https://travis-ci.org/deis/helm set up to build PRs with smarts (only test charts that have been modified in the PR).

I'll wait to close this issue until we actually are doing the `helm install && helm uninstall` actions

technosophos:
Status?

sgoings:
Closing in preference of helm/charts#43

PR #15, opened 2015-10-21
feat(test): add ci test infrastructure scaffolding

sgoings:
Currently the makefile needs some love in other places (why it's commented out inside of the `make test-charts` logic), but this is my idea behind how we test PRs for charts:

If it's a PR, we see what charts have changed and then execute 

```
helm install <chart>
helm test <chart>
helm uninstall <chart>
```

(example: https://travis-ci.org/sgoings/helm/builds/86681185)

If it's master we just run a set of default chart installs (overrideable via TEST_CHARTS env)

Related #4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,1 @@
+script: make test-charts
--- a/Makefile
+++ b/Makefile
@@ -36,5 +36,7 @@ dist: build-all
 	$(DIST_DIRS) zip -r helm-{}.zip {} \; && \
 	cd ..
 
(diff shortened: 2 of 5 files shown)


technosophos:
LGTM. I think we'll go with one LGTM until we reach a stable point. So feel free to merge.

sgoings:
I realized that having a `helm test` doesn't make as much sense as just running a test as a part of the install. What do you think?

I'm thinking just running:
```
helm install
helm uninstall
```

PR #29, opened 2015-10-21
feat(*): add basic install

technosophos:
This adds a very basic version of install. We still need to make the actual call to Kubectl a little more intelligent, but this will be useful for testing.


gabrtv:
LGTM.

PR #31, opened 2015-10-21
update LF address

dwmcallister:
changed to current Linux Foundation address


technosophos:
Merging. Thanks @dwmcallister

PR #35, opened 2015-10-22
feat(target) add helm target command

michelleN:


helm/action/target.go
```diff
@@ -0,0 +1,17 @@
+package action
+
+import (
+	"fmt"
+	"os/exec"
+
+	"github.com/deis/helm/helm/log"
+)
+
+func Target() {
```
mboersma: An exported func should have a doc string.

technosophos:
LGTM

michelleN:
This takes care of https://github.com/deis/helm/issues/18

PR #37, opened 2015-10-22
docs(README.md): clarify that helm is a WIP

mboersma:
@technosophos suggested a logo to indicate that certain deis/* umbrella projects are considered in progress and don't necessarily follow the same process as deis/deis. Here is a simple one.


michelleN:
:+1:

PR #45, opened 2015-10-22
Info

adamreese:
https://github.com/deis/helm/issues/12

Working but still needs some assertions

### DIFF
--- a/helm/action/info.go
+++ b/helm/action/info.go
@@ -1,0 +1,27 @@
+package action
+
+import (
+	"path/filepath"
+
+	"github.com/deis/helm/helm/log"
+	"github.com/deis/helm/helm/model"
+)
+
+func Info(chart, homedir string) {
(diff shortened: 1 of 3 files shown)


helm/action/info.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"path/filepath"
+
+	"github.com/deis/helm/helm/log"
+	"github.com/deis/helm/helm/model"
+)
+
+func Info(chart, homedir string) {
+	chartPath := filepath.Join(homedir, "cache/charts", chart, "Chart.yaml")
```
technosophos: Minor: This should use the constant CacheChartPath instead of "cache/charts".

technosophos:
Only that one minor change, and then LGTM.

helm/action/info.go
adamreese: Didn't see that const.  Should we move global const's to a `action.go`

PR #48, opened 2015-10-23
feat(*): add object model for charts

technosophos:
- Add object model
- Add lots of test data
- Add YAML processing API on top of k8s

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ test:
 
 clean:
 	rm -f ./helm/helm.test
-	rm -f ./helm
+	rm -f ./helm.bin
 
 bootstrap:
 	glide -y glide-full.yaml up
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,9 @@ install: build
 test:
 	go test -v ./helm/. ./helm/manifest ./helm/action ./helm/log ./helm/model
 
+quicktest:
+	go test ./helm/. ./helm/manifest ./helm/action ./helm/log ./helm/model
+
 clean:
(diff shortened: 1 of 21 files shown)


technosophos:
This is failing because it needs `glide -y glide-full.yaml up`, which is still unmerged (#49).

mboersma:
After the `glide -y` change, this seems to fail still:
```console
go build -o helm.bin -ldflags "-X main.version=" helm/helm.go
vendor/k8s.io/kubernetes/pkg/api/resource/quantity.go:27:2: cannot find package "speter.net/go/exp/math/dec/inf" in any of:
	/home/travis/gopath/src/github.com/deis/helm/vendor/speter.net/go/exp/math/dec/inf (vendor tree)
	/home/travis/.gimme/versions/go1.5.1.linux.amd64/src/speter.net/go/exp/math/dec/inf (from $GOROOT)
	/home/travis/gopath/src/speter.net/go/exp/math/dec/inf (from $GOPATH)
make: *** [build] Error 1
```

technosophos:
Oh, geck... that package moved. k8s is fetching it straight off of a website instead of using gopkg.in. I filed an upstream bug in Glide about that. I guess I should file one with k8s too.

For now, I'll set up the makefile to fetch it with `go get`.

technosophos:
I talked to @mattfarina, and he says that the latest VCS package fixes this. So we should just build Glide from `master` for the time being. Right now @sgoings is working on the Travis config. I paid him in chocolate.

gabrtv:
LGTM.

PR #53, opened 2015-10-23
fix(*): updated README and changed clean

technosophos:
README now notes that a Go developer needs Glide. And the Makefile
clean target now removes the right binary.

Addresses #52

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,11 +18,15 @@ A Chart is a unit of Kubernetes manifests that reflect best practices as determi
 
 ## Install Helm
 
-To just use Helm, you can run the following:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,10 +18,11 @@ A Chart is a unit of Kubernetes manifests that reflect best practices as determi
 
 ## Install Helm
 
+To just use Helm, you can run the following:
```
mboersma: Right below this is "Using Helm" so I found this wording confusing. Maybe:
```
To install Helm, run this command:
```
nitpick! :green_book:

technosophos:
Fixed up the install instructions per @mboersma 's comment.

PR #58, opened 2015-10-23
test(info): add unit test for info command

adamreese:


adamreese:
Requires https://github.com/deis/helm/pull/67

adamreese:
closing in favor of https://github.com/deis/helm/pull/112

PR #63, opened 2015-10-23
doc: add a document explaining the workspace directory

technosophos:
Based on a conversation with @gabrtv I wrote up this document explaining how the workspace is designed.


gabrtv:
Terrific write-up @technosophos.

PR #64, opened 2015-10-23
fix: check number of arguments and print an useful error

technosophos:
This implements a simple argv check that prints usage when not enough arguments are specified. I can't seem to find a way to do this automatically with codegangsta/cli.


technosophos:
Oh, my bad... I pushed this to deis/helm instead of technosophos/helm. Sorry. I rebased and re-pushed though.

PR #65, opened 2015-10-23
Colorize the output of helm commands

technosophos:
Addresses #5


technosophos:
I'm going to leave this one for a bit while I sort out some of the deeper functionality. Right now, it requires `deis/pkg`, which in turn will suck in both etcd and k8s. Not sure I want that.

technosophos:
This really is ready for review now.

PR #74, opened 2015-10-26
chore(log): set debug with command flag

adamreese:


technosophos:
Nice! I knew there must be some way of doing that!

PR #85, opened 2015-10-27
feature(update) show local and remote info on update

slack:
Give a bit more info when running update.


slack:
```
helm [update-info]$ helm update
Updated /Users/jhansen/.helm/cache from https://github.com/deis/helm
helm [update-info]$
```

PR #87, opened 2015-10-27
doc(skel) make the skeleton Chart generic

slack:
Make the skeleton more generic. Prompt for the things we would like to see.

Might also be a good opportunity to expand the skeleton to include a Service definition, so we can explain our approach to labels in situ.

### DIFF
--- a/skel/manifests/example-pod.yaml
+++ b/skel/manifests/example-pod.yaml
@@ -1,10 +1,10 @@
 apiVersion: v1
 kind: Pod
 metadata:
-  name: alpine
+  name: example-pod
 spec:
   restartPolicy: Never
   containers:
-  - name: alpine
+  - name: example
   image: "alpine:3.2"
     command: ["/bin/sleep","9000"]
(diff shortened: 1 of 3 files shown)


skel/manifests/example-pod.yaml
```diff
@@ -1,10 +1,10 @@
...
 kind: Pod
 metadata:
-  name: alpine
+  name: example-pod
 spec:
   restartPolicy: Never
   containers:
-  - name: alpine
+  - name: example
   image: "alpine:3.2"
```
slack: One question is should the skeleton chart actually "work"? Debated about setting this to `scratch` and a random command.

technosophos: I don't know if that's a terribly big deal, but it certainly wouldn't hurt. That said... could you _do_ "scratch" and a random command? I don't think scratch has any commands? Maybe "alpine:3.2" with something like "sleep 900" or something?

slack: `scratch` + command wouldn't work. To help prompt folks writing a chart, we may  as well leave this as something that does something, even if it is sleep.

slack:
Rebased again.

slack:
Rebased.

PR #103, opened 2015-10-27
feat(helm): add --dry-run flag to install

### LINKED ISSUE #78: Add --dry-run to helm install
This option should show all the output that would be sent to k8s, but not actually run kubectl.

bacongobbler:
closes #78 

This is my first time contributing to Helm. Woo! Anything else I should add here? docs/tests?

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ Some of the features we plan to tackle in the near future:
 - [ ] Mandatory Group Labels [#80](https://github.com/deis/helm/issues/80)
 - [ ] Linting for Charts [#96](https://github.com/deis/helm/issues/96)
 - [ ] End-to-End Testing of Charts [#4](https://github.com/deis/helm/issues/4)
-- [ ] Dry-Run Installs [#78](https://github.com/deis/helm/issues/78)
+- [x] Dry-Run Installs [#78](https://github.com/deis/helm/issues/78)
 - [ ] Improved Dependency Resolution (based on service provider)
 - [ ] Upgrading Charts
 
--- a/action/install.go
+++ b/action/install.go
@@ -1,6 +1,7 @@
 package action
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -26,7 +27,7 @@ import (
 // 	- Services
 // 	- Pods
 // 	- ReplicationControllers
-func Install(chartName, home, namespace string, force bool) {
+func Install(chartName, home, namespace string, force bool, dryRun bool) {
 	if !chartFetched(chartName, home) {
 		log.Info("No chart named %q in your workspace. Fetching now.", chartName)
 		fetch(chartName, chartName, home)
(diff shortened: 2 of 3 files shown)


helm/helm.go
```diff
@@ -81,6 +81,10 @@ func main() {
 					Name:  "force, aye-aye",
 					Usage: "Perform install even if dependencies are unsatisfied.",
 				},
+				cli.BoolFlag{
+					Name:  "dry-run",
+					Usage: "Run but do not install into Kubernetes.",
```
sgoings: Run vs install is confusing verbiage to me... Maybe something like: "show but do not execute kubectl commands"?

bacongobbler: sounds good to me :)

technosophos: I think what @bacongobbler was getting at in help is that this will do the `fetch` part if necessary, but will not run the `kubectl` part. We should probably make that clear.

I'm wondering if it makes sense to pass the dryRun flag into `uploadManifests` and have it print out what it would normally do. That would give the user a chance to see whether the final generated JSON is what they anticipate.

Thoughts, @sgoings ?

bacongobbler: That would be good to display. We could just print the `kubectl` command on stdout when `uploadManifests` is called.

Does "perform the fetch, but only show kubectl commands" sound like a good docstring? Or is that not the intended behaviour we want?

sgoings: That wording is still missing the part about skipping the actual kubectl execution...

technosophos: @sgoings I was just thinking of what many other programs do during _dry run_ operations. It feels like we'd want to step as far through the process as possible without actually making any changes to Kubernetes.

bacongobbler:
Done!

```
$ ./bin/helm help install
NAME:
   ./bin/helm install - Install a named package into Kubernetes.

USAGE:
   ./bin/helm install [command options] [chart-name...]

DESCRIPTION:
   If the given 'chart-name' is present in your workspace, it
will be uploaded into Kubernetes. If no chart named 'chart-name' is found in
your workspace, Helm will look for a chart with that name, install it into the
workspace, and then immediately upload it to Kubernetes.

When multiple charts are specified, Helm will attempt to install all of them,
following the resolution process described above.

As a special case, if the flag --chart-path is specified, Helm will look for a
Chart.yaml file and manifests/ directory at that path, and will install that
chart if found. In this case, you may not specify multiple charts at once.


OPTIONS:
   --chart-path, -p 	An alternate path to fetch a chart. If specified, Helm will ignore the cache.
   --namespace, -n 	The Kubernetes destination namespace.
   --force, --aye-aye	Perform install even if dependencies are unsatisfied.
   --dry-run		Fetch the chart, but only display the underlying kubectl commands.

$ ./bin/helm install --dry-run alpine
---> Performing a dry run of `kubectl create -f` ...
---> kubectl create -f - < {"kind":"Pod","apiVersion":"v1","metadata":{"name":"alpine","creationTimestamp":null,"annotations":{"HelmOriginFile":"/Users/bacongobbler/.helm/workspace/charts/alpine/manifests/alpine-pod.yaml"}},"spec":{"containers":[{"name":"waiter","image":"alpine:3.2","command":["/bin/sleep","9000"],"resources":{}}],"restartPolicy":"Never"},"status":{}}
---> Done
========================================
# Alpine

This Chart provides a Pod running Alpine Linux.
========================================
$ k get pods
NAME        READY     STATUS    RESTARTS   AGE
```

bacongobbler:
The tab spacing is gross in the flags. Is that a bug with codegangsta/cli or something we can fix up?

technosophos:
Thank you for making these changes, @bacongobbler! This is exactly what I was envisioning when I asked for the changes.

gabrtv:
@bacongobbler can you please add a strikethrough in the root readme for the `--dry-run` flag under the Future Plans section?

bacongobbler:
@gabrtv it is done :)

PR #104, opened 2015-10-27
feat(search): search chartfile name and description

### LINKED ISSUE #11: Better `helm search`
The current `helm search` only looks at the names of charts.

Could we extend to be a little bit better? Maybe search the description?

Things that _might_ be useful to have searched:

- Chart.yaml...
  - name
  - description
- manifests/
  - metadata/labels headers
  - `image:` field (which shows docker container images by name)

lanej:
* Fixes #11

### DIFF
--- a/helm/action/search.go
+++ b/helm/action/search.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"path/filepath"
+	"regexp"
 
 	"github.com/deis/helm/helm/log"
 	"github.com/deis/helm/helm/model"
@@ -11,8 +12,7 @@ import (
 
 // Search looks for packages with 'term' in their name.
 func Search(term, homedir string) {
-
-	dirs, err := search(term, homedir)
+	charts, err := search(term, homedir)
 	if err != nil {
 		log.Die(err.Error())
 	}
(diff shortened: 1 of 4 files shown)


helm/action/list.go
```diff
@@ -10,13 +10,15 @@ import (
 // List lists all of the local charts.
 func List(homedir string) {
 	md := filepath.Join(homedir, WorkspaceChartPath, "*")
+
 	charts, err := filepath.Glob(md)
-	if err != nil {
```
carmstrong: I think the paradigm in go is `err != nil`... any reason for this change?

carmstrong:
Good stuff, @lanej. Can you fix up the commit subjects? Squashing them into one titled `feat(search): deeper search` or something like that should be good.

lanej:
ah yes, the ole commit message format.

helm/action/list.go
lanej: just like to prevent accidental assignments instead of comparison.  it's a habit. i can back this out.

technosophos: @carmstrong is right that the Go idiom is always to do `varname != nil`. Go will catch the case of accidental assignment.

technosophos:
Other than switching back to the idiomatic `err != nil` in several places, I'm good with this.

Looks like if no argument is passed in, the `""` will be used.

lanej:
@technosophos good to know. ill back those out.

carmstrong:
@technosophos LGTY now?

technosophos:
I am totally good with this now. Thank you very much @lanej!

PR #107, opened 2015-10-28
[WIP] feat(*): replace k8s with kubelite

technosophos:
This branch is an experiment in removing the Kubernetes library from Helm's dependencies.


sgoings:
Are we still taking this road @technosophos? My vote: I don't think we should... and instead just show the k8s community our struggles. :smile:

technosophos:
I took a few parts of the YAML encoding and moved them out to a separate commit. I'm closing this one, since I think it's too broad. Most developers seem to be able to get over the initial install hurdle.

PR #112, opened 2015-10-28
feat(info): add optional custom format

adamreese:
Allow for custom formatting of `helm info` using a Go template

Replaces: https://github.com/deis/helm/pull/58


technosophos:
The tests aren't passing

carmstrong:
This also needs a rebase now, @adamreese

adamreese:
@technosophos it's an issue of other unit tests changing the current working directory.  I have a fix in progress

helm/action/helper_test.go
```diff
@@ -0,0 +1,36 @@
...
+	// create cache directory
+	tmpHomeCache := filepath.Join(tmpHome, "cache")
+	os.Mkdir(tmpHomeCache, dstat.Mode())
+
+	// absolute path to testdata charts
+	testChartsPath := filepath.Join(helmRoot, "helm/testdata/charts")
+
+	// copy testdata charts into cache
+	// mock git clone
+	exec.Command("cp", "-R", testChartsPath, tmpHomeCache).Run()
```
lanej: no golang lib function for this ?

adamreese: noop, you need to walk the dir and copy each file

technosophos: There's a copyDir() function in (I think) `action/update.go` that does essentially this.

technosophos:
I'd prefer using the `copyDir()` func if it suits your needs, but I'm going to go ahead and LGTM this.

helm/action/helper_test.go
adamreese: oh I missed that, thanks

PR #113, opened 2015-10-28
docs(README): add Go Report Card badge

mattfarina:


bacongobbler:
+1

mboersma:
Helm gets an A+ on 31 files. Nice! We should add linters to the Makefile in `make test-style` as we do in deis/deis.

Issue #116, opened 2015-10-28
Update README


smugcloud:
Since the etcd chart does not currently exist, we should remove it to help new users get up and running.

PR #117, opened 2015-10-28
updated readme for #116

smugcloud:
Addresses [#116] (https://github.com/deis/helm/issues/116)


bacongobbler:
works for me! Would you mind amending your commit so that it follows our [commit style guide](http://docs.deis.io/en/latest/contributing/standards/#commit-style), pretty please? That way our changelog generation script will be able to catch your commit.

Something like `fix(readme): update helm search usage` should do the trick. Thanks! :)

Also this needs a rebase :heart:

michelleN:
Looks good to me as well. I might suggest adding the `helm edit example-todo` line in there so that users know they can edit a chart as well.

sgoings:
We should probably fix the ".git" and README.md from showing up in a `helm search`...

gabrtv:
@sgoings agreed, though we should probably merge as-is and then fix output when we fix the `helm search` behavior. This already has 2 LGTMs.

gabrtv:
I'm going to fix the client output and update the readme in that PR.  Closing.

Issue #55, opened 2015-10-23
`helm remove` to remove a chart from the workspace


technosophos:
While `helm fetch` copies a chart into the workspace, `helm remove` should remove the chart. It should probably have the alias `helm rm`.

technosophos:
@michelleN would you like to take this one since I swiped `helm uninstall` out from under you? (Again, sorry... my bad)

michelleN:
haha yeah I'll take it. no worries. thanks!

PR #119, opened 2015-10-29
feat(remove) helm remove command

michelleN:
`helm remove [chart]` removes a chart from the
workspace directory. This resolves issue #55.

### DIFF
--- a/helm/action/remove.go
+++ b/helm/action/remove.go
@@ -1,0 +1,25 @@
+package action
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/deis/helm/helm/log"
+)
+
+// Remove removes a chart from the workdir.
+//
+// - chart is the source
+// - homedir is the home directory for the user
+func Remove(chart string, homedir string) {
+	chartPath := filepath.Join(homedir, WorkspaceChartPath, chart)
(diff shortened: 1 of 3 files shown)


michelleN:
@technosophos - yeah I don't think we can test this without messing up testdata/. The better solution would be to maybe copy the testdata directory as part of the test run, set the copy as the home directory and then call the remove function to see if the chart was removed. I believe this was @adamreese's  plan so I'm going to work on that with him. What do you think?

adamreese:
@michelleN I stuffed it in this https://github.com/deis/helm/pull/112

michelleN:
sweet. yeah, just saw that. @adamreese

michelleN:
Thanks for pointing out the go idiom @technosophos. I changed it.

technosophos:
Okay, this is looking good. Should I LGTM?

michelleN:
sure @technosophos

michelleN:
@technosophos I've added some tests on another branch that uses @adamreese's testing helper functions but I can merge those later.

michelleN:
Added some tests. Now it's ready for some reviewing/lgtm-ing. :dancer:

Issue #124, opened 2015-10-29
glide version issues


michelleN:
@lanej and I ran into some issues related to the version of glide. Turns out we needed HEAD of glide and the glide installed using brew was not at HEAD. To fix the issue, we ran `brew uninstall glide` and ran ` go get github.com/Masterminds/glide`. 
This did the trick for us, so I figured if these instructions were in the README, it might help helm users by saving them some time.

PR #125, opened 2015-10-29
docs(README): add glide version warning

michelleN:
This resolves issue #124.


bacongobbler:
Alternatively, if you want to build from a tagged version of `glide`:

```
cd glide
git checkout 0.6.1
GO15VENDOREXPERIMENT=1 make bootstrap build
sudo make install
```

using `go get` just makes it a bit easier, too.

slack:
LGTM, we should come back to this when the next version of glide is released and recommend >= 0.6.2.

michelleN:
sure thing @slack

PR #133, opened 2015-11-02
fix(tests): reset log after test run

adamreese:


gabrtv:
This is probably better addressed with proper setup/teardown, but LGTM for now. ping @mboersma

mboersma:
>  probably better addressed with proper setup/teardown

Right, now we can use `TestMain` to run traditional setup/teardown logic for each test case. Here's an example from a deis test refactoring branch I have:
```go
// TestMain performs global setup, runs the tests, and cleans up after tests. This function is
// called by `go test` instead of running the tests directly, as of Go 1.4.
func TestMain(m *testing.M) {
	setup()
	retCode := m.Run()
	teardown()
	os.Exit(retCode)
}

// setup creates the common test environment.
func setup() {
}

// teardown cleans up after testing.
func teardown() {
}
```

(This PR's approach is totally fine with me, this is just to demonstrate what @gabrtv was referring to.)

adamreese:
oh that is cool, I will look into using that

PR #137, opened 2015-11-02
fix(create): remove skel directory

adamreese:
Fixes `helm create` erroring out due to missing resources

* Embed skel resources
* Create directory structure if missing
* Populate Chartfile and Readme with chartname

Fixes https://github.com/deis/helm/issues/130


michelleN:
:+1:

gabrtv:
This approach generally looks good and should be easier to maintain over time, but deferring to @technosophos.

helm/action/create.go
```diff
@@ -1,33 +1,100 @@
...
-	chartDir := path.Join(homeDir, WorkspaceChartPath, chartName)
+	log.Info("Created chart in %s", chartDir)
+}
+
+// newSkelChartfile populates a Chartfile model with example data
+func newSkelChartfile(chartName string) *model.Chartfile {
+	return &model.Chartfile{
+		Name:        chartName,
+		Home:        "http://example.com/your/project/home",
+		Version:     "0.0.1",
```
bacongobbler: let's change the version number to 0.1.0. According to semver, 0.1.0 is a minor release, whereas 0.0.1 is a patch release of 0.0.0 which doesn't make that much sense :)

helm/action/create.go
```diff
@@ -1,33 +1,100 @@
...
+Describe your chart here. Link to upstream repositories, Docker images or any
+external documentation.
+
+If your application requires any specific configuration like Secrets, you may
+include that information here.
+`
+
+// manifestSkel is an example manifest for a new chart
+const manifestSkel = `apiVersion: v1
+    kind: Pod
```
bacongobbler: since this is a raw string literal, we probably want this without the extra spacing at the start so it's all nicely formatted yaml.

adamreese:
@technosophos please check me on the template and lable

technosophos:
Yup, those look good @adamreese.

@bacongobbler pointed out some indention issues, and I am actually wondering if the YAML template for the pod will be broken by default if it is indented the way it currently is. Haven't tested though.

helm/action/create.go
adamreese: @bacongobbler Updated the formatting on the yaml

helm/action/create.go
adamreese: @bacongobbler if you patch `nil` is it still `nil`? :smile:

Issue #92, opened 2015-10-27
Rename skel, test, and docs to _skel, _test, and _docs.


technosophos:
Per #90, we will move the source code of helm up a directory. To better accomodate the Go tools, we should rename the above directories. (Go ignores directories prefixed with `_`). That way, if we end up putting any Go-like code in these directories, the Go runtime will ignore them.

bacongobbler:
since we're breaking out all of the external libs into places like deis/pkg, wouldn't it make more sense to store the source code in something like

michelleN:
@bacongobbler - I also like that project structure. If we went that route, would there be any additional complexities using go tools with the helm project that you might know of? @technosophos any thoughts?

bacongobbler:
There would be a bit more internal complexity compared to your run-of-the-mill Go project, but I believe the benefits outwiegh the costs.

For starters, we'd have to ensure that the $GOPATH is set to $GOPATH/src/github.com/deis/helm when running a build, which some go tools would pick up automatically ([gb](https://github.com/constabulary/gb) does this by recursively searching for a `src/` directory).

One project that immediately comes to mind that uses this project structure would be [confd](https://github.com/kelseyhightower/confd). I help maintain that project though so I may be a bit biased, but I believe that project structure works quite well.

For example, the project structure (in my mind) would look something like this:

```
etc/helm/skel/
docs/
tests/
tests/ci/
src/
  helm.go
  helm/
    model/
    manifest/
    [...]
```

In helm.go, importing helm libraries would no longer be `github.com/deis/helm/helm/model`, but instead just `helm/model`. `glide` would bootstrap vendored dependencies into `./src/vendor`, which then Go with the 1.5 vendor experiment would work as normal: 

```
$ cd helm
$ make bootstrap
$ GOPATH=$PWD go build src/...
```

In terms of benefits to this project structure, forking and contributing to the project just got a heck of a lot easier. The user doesn't have to import the project into $GOPATH/src/github/com/deis/helm to build the project, which has burned us in the past with deis who are new to Go. More experienced Go developers may be ticked off that we're not using Go's native tooling "The Way It Was Meant To Be:tm:", but we're already using `glide` so they already have something else to gripe about ;)

One cost of this route means that the project can no longer be used as a library, but we can always split those out of the project like we do with [pkg](https://github.com/deis/pkg).

If it was up to me, this is the project structure that I would tinker with. :)

michelleN:
[...]

PR #138, opened 2015-11-02
chore(*) rename test/ to  _test/

michelleN:
This resolves issue #92.


technosophos:
This looks good to me. IIRC, @bacongobbler had a reason to not rename `docs` to `_docs`, but when I read back through the issue, I couldn't find it.

michelleN:
@technosophos could be for mkdocs reasons? @bacongobbler

bacongobbler:
yup. I don't think we'll ever be writing .go files in docs/ so I didn't think it made sense to add an underscore to that dir.

michelleN:
okay cool I'll take that out.

michelleN:
ready for review @technosophos @bacongobbler

Issue #136, opened 2015-11-02
Notify user of update to Helm


technosophos:
As a user, I would like to be notified when a newer stable version of Helm is available.

In the last hour, both @gabrtv and @jackfrancis have asked about this, and it sure would be nice to have in the very first release.

One possible (and cheap) solution would be to have Helm query the Releases available for the Helm repo, and do a simple tag-based match to see if it is on that release.

We'll shoot for something like this:

```
$ helm update
---> A newer version of Helm is available. Download version 1.2.3: http://github.com/deis/helm/...
---> # Rest of the normal output
```

For this issue, it'll be a non-goal to actually install a newer version, though we should probably look into this in the future.

rimusz:
+1

PR #140, opened 2015-11-02
feat(*): show information when an updated Helm version is available

technosophos:
When a newer release is indicated on the Release page of GitHub, this will display version information.

It works by comparing the remote version with the local version, and suggesting an update. This is totally SemVer 2.0 based.

Addresses #136


adamreese:
We should add a test for this soon...

technosophos:
I guess we could mock the GitHub library and simulate some versions. That actually wouldn't be too hard.

bacongobbler:
We could also do a standalone test where we compile helm with version = latest - 1. That way we're not relying on mocks, but that means we rely on github uptime.

carmstrong:
:+1: For testing the function

bacongobbler:
LGTM (on mobile). Nice test.

carmstrong:
:ship:

PR #147, opened 2015-11-03
fix(Makefile): create dist dir automatically

sgoings:


mattfarina:
Should the directory be `dist` or `_dist`? I noticed the non-go related directories now have an _ prefix.

sgoings:
Good call @mattfarina. I will update.

bacongobbler:
good call. I would've missed that.

PR #156, opened 2015-11-03
fix(create): fix yaml formatting in skeleton manifest

gabrtv:
This fixes a bug in YAML formatting that would manifest during:

```
$ helm create mychart
$ helm install mychart
<yaml error>
```

### DIFF
--- a/action/create.go
+++ b/action/create.go
@@ -31,7 +31,7 @@ spec:
   restartPolicy: Never
   containers:
   - name: example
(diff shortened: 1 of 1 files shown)


adamreese:
thanks for fixing my bug .__.

gabrtv:
No worries. Once @michelleN gets a linter in place, we should probably backfill some unit tests.

PR #160, opened 2015-11-03
fix(.gitignore): hide unit test droppings

mboersma:
Running `make test` creates a testdata/helm_home/ directory which causes git to think the repository has uncommited files. I also alpha-sorted the .gitignore entries, because I could not help myself.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
-vendor/
-bin/
 _dist/
+bin/
 ci/bintray-ci.json
-
+testdata/helm_home/
+vendor/


technosophos:
This used to be in `.gitignore` until @adamreese removed it (check the history). We're playing ping-pong here.... Can we get some agreement on whether or not this should be ignored?

technosophos:
My vote, incidentally, is that that directory should be ignored. It is supposed to be created during the `update` test. That test should probably be fixed to remove its own droppings.

mboersma:
Ah yes, how quickly we (I) forget. Well the `helm/testdata/helm_home/cache` file from #58 is not now in the repository, and the unit tests do pass without that directory existing, which was the motivation for #67. So it looks to me like this now should be ignored, but we should see what @adamreese thinks.

adamreese:
@mboersma the `.gitignore` file was revised to get past bad setup/tear down of tests.  The only test not updated was the `helm update` test.  I would say fix the test so it is repeatable locally and on CI and leave the `.gitignore` as a warning that something was not cleaned up.

technosophos:
That particular test was supposed to create those directories. It probably just needs to remove the directories when it's done.

adamreese:
Yep, or use a temp directory for home

adamreese:
*thread-safety first!*

mboersma:
I see--so there is still a `helm/testdata/helm_home/` in version control, but `make test` also creates `testdata/helm_home`... This probably isn't the way to fix it properly; I'll create an issue: #164.

Issue #118, opened 2015-10-29
PROPOSAL: Chart tables for Adding Custom (Public or Private) Repos


technosophos:
In nautical terminology, a Chart Table is a place where charts are
stored. This proposal outlines a way to extend Helm's notion of charts
to include arbitrary external sources, which we call Chart Tables.

**Changes:**
- Change `charttable` to `table`.
- Added Namespacing section
- Changed Charttables.yaml to Tables.yaml

## TL;DR

Add a method for supplying external "chart tables" (chart repositories),
and seemlessly use them within the tool.

```
$ helm table list
 charts (default)
$ helm table add mycharts https://github.com/myname/myrepo
----> Added new repo named 'myrepo'
$ helm table list
 charts (default)
 mycharts
$ helm update
----> Updated 'charts'
----> Updated 'myrepo'
$ helm search myrepo/*
 mychart1
 mychart2
$ helm fetch myrepo/mychart2
----> Installing 'myrepo/mychart2` into $HELM_HOME/workspace
....
```

## Goals

- Keep a base repository where many prominent charts can go
- Allow users to add _ad hoc_ public Git-based repositories
- Support private repositories, too!

Non-goal: We are not trying to switch to a fully distributed
chart-per-repo model.

## What Is A Chart Table?

A chart table is an extra source of charts. The default chart table, which
is built into Helm, is `github.com/deis/charts`. This spec defines how
additional chart tables are added, stored, and utilized.

To build a chart table, we need three pieces of information:

- the local name of the chart table
- the remote access point
- the protocol(s) used to access the remote

At this point, we suggest that (a) all remote access points are Git
repos, and (b) only git protocols are suppoted.

## The Structure of the $HELM_HOME Chart Tables

Inside of `$HELM_HOME`, charts are currently stored inside of
`cache/charts`. The `cache` directory has no other subdirectories. The
reasoning for having it this way is so that there is a place to put
extra chart tables. So we can carry out the initial design by adding
structure as follows:

```
$HELM_HOME
    |
    |- cache/
        |
        |- charts/    # Default charts
        |
        |- CHARTTABLE1/
        |
        |- CHARTTABLE2/
```

In the above, each chart table (`charts`, `CHARTTABLE1`, `CHARTTABLE2`) contains
any number of charts.

### Namespace Model

The chart tables are organized into namespaces of the form `tablename/chartname`. 

Therefore, to install a chart named `nginx` in a table named `jackfrancis`, the proper install command is:

```
[...]

Issue #111, opened 2015-10-28
Support multiple repos within helm cache directory


sgoings:
`helm`'s current behavior in using 2 repositories is:

```
~/go/src/github.com/deis/helm  ⑂ master +    rm -rf ~/.helm/

 ~/go/src/github.com/deis/helm  ⑂ master +    ./bin/helm update
---> Creating /Users/sethgoings/.helm
---> Creating /Users/sethgoings/.helm/cache
---> Creating /Users/sethgoings/.helm/workspace
---> Creating /Users/sethgoings/.helm/cache/charts
---> Cloning repo into "/Users/sethgoings/.helm/cache/charts". Please wait.
---> Updated /Users/sethgoings/.helm/cache/charts from https://github.com/deis/charts

 ~/go/src/github.com/deis/helm  ⑂ master +    ./bin/helm --repo https://github.com/bacongobbler/charts.git update
[ERROR] Could not get repository "https://github.com/bacongobbler/charts.git": The Remote does not match the VCS endpoint

 ~/go/src/github.com/deis/helm  ⑂ master +    rm -rf ~/.helm/

 ~/go/src/github.com/deis/helm  ⑂ master +    ./bin/helm --repo https://github.com/bacongobbler/charts.git update
---> Creating /Users/sethgoings/.helm
---> Creating /Users/sethgoings/.helm/cache
---> Creating /Users/sethgoings/.helm/workspace
---> Creating /Users/sethgoings/.helm/cache/charts
---> Cloning repo into "/Users/sethgoings/.helm/cache/charts". Please wait.
---> Updated /Users/sethgoings/.helm/cache/charts from https://github.com/bacongobbler/charts.git
```

It would be nice if helm could support multiple repositories within the cache dir. Additionally, it might be good if `helm update` updated **all** repositories within cache. And a `helm update <repo name>` would update just an individual repo within cache.

Issue #109, opened 2015-10-28
support discovery of Charts via other repositories for folks creating non-open source Charts


jstrachan:
up to now we've been releasing the versioned kubernetes resources for running various apps in the [fabric8 project](http://fabric8.io/) using maven repositories as a canonical place to find them. We went with maven repositories as lots of folks we work with use upstream OSS code and have on premise stuff and need to work with both kinds of resources and lots of these folks already have maven repositories.

e.g. here you can search for them in maven centrals repository for many of fabric8's kubernetes resources:
http://search.maven.org/#search%7Cga%7C1%7Cl%3A%22kubernetes%22

right now we've been releasing OpenShift template JSON; but we'll happily generate helm Charts too. It would be nice if there was some mechanism to discover Charts in other repositories though.

e.g. I can imagine folks wanting to use a helm-like tool for internal Charts which are not public on the internet and so can't be submitted as a PR on the public helms repository.

I guess an easier option than supporting maven repository searching would be to support multiple git repositories for the charts repositories; so there can be the main public OSS helm git repo; then users can have their own internal git repos?

bacongobbler:
I think that would be a good feature. `pip` and other tools allow communicating with another index other than pypi.python.org, and `brew` has `brew tap`. Having the ability to pick and choose repositories to install from like `brew tap` would be a good feature to have at some point.

sgoings:
I believe we already have the beginnings of support for multiple git repos via:

```
GLOBAL OPTIONS:
   --repo "https://github.com/deis/helm"	The remote Git repository as an HTTP URL [$HELM_REPO_URL]
```

Example:
```
./bin/helm --repo https://github.com/bacongobbler/charts update
---> Creating /Users/sethgoings/.helm
---> Creating /Users/sethgoings/.helm/cache
---> Creating /Users/sethgoings/.helm/workspace
---> Cloning repo into "/Users/sethgoings/.helm/cache". Please wait.
---> Updated /Users/sethgoings/.helm/cache from https://github.com/bacongobbler/charts
```

bacongobbler:
see also #111

PR #162, opened 2015-11-03
feat(*): add repo commands

technosophos:
This adds the following commands:

```
helm repo add [lname] [gitURL]
helm repo rm [lname]
helm repo list
```

This addresses #118 #111 and #109.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,10 +47,10 @@ else
 endif
 
 quicktest:
-	go test ./. ./manifest ./action ./log ./chart ./dependency
+	go test ./. ./manifest ./action ./log ./chart ./dependency ./repo
 
 test:
-	go test -v ./. ./manifest ./action ./log ./chart ./dependency
+	go test -v ./. ./manifest ./action ./log ./chart ./dependency ./repo
 
 test-charts:
 	@./_test/test-charts $(TEST_CHARTS)
--- a/action/search.go
+++ b/action/search.go
@@ -3,9 +3,9 @@ package action
 import (
 	"errors"
 	"fmt"
-	"os"
 	"path/filepath"
 	"regexp"
+	"sort"
 
 	"github.com/deis/helm/chart"
 	"github.com/deis/helm/log"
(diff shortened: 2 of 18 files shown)


action/search.go
```diff
@@ -13,52 +13,70 @@ import (
...
+	}
 
 	for _, dir := range dirs {
-		chart, err := chart.LoadChartfile(filepath.Join(dir, "Chart.yaml"))
+		cname := filepath.Join(table, filepath.Base(dir))
+		chrt, err := chart.LoadChartfile(filepath.Join(dir, "Chart.yaml"))
 
 		if err != nil {
-			log.Warn("failed to load Chart.yaml: %v", err)
+			//log.Warn("Failed to load Chart.yaml: %v", err)
```
mboersma: It's ok not to log the error here?

Makefile
```diff
@@ -47,10 +47,10 @@ else
 endif
 
 quicktest:
-	go test ./. ./manifest ./action ./log ./chart ./dependency
+	go test ./. ./manifest ./action ./log ./chart./dependency ./repo
```
adamreese: need a space `./chart ./dependency`

action/search.go
technosophos: Right. It is not an error for a directory to not be a `chart`. This was generating a lot of spurious errors for directories like `.git`. (I actually removed the log message in one of the other PRs somewhere in here... so even the comment will go away)

adamreese:
@technosophos what is the thought behind a `Repofile`?  A generic helm config file would scale better for new features that require persistent settings

technosophos:
Fixed `quicktest` and replaced the commented log message with a comment explaining why we intentionally skip.

mboersma:
Groovy:
```console
$ helm up
---> Creating /Users/matt/.helm/cache/Repos.yaml
---> Updating charts
---> Done
$ helm repo add bacongobbler https://github.com/bacongobbler/charts
---> Cloning into '/Users/matt/.helm/cache/bacongobbler'...
$ helm repo list
	charts*	https://github.com/deis/charts
	bacongobbler	https://github.com/bacongobbler/charts
$ helm fetch bacongobbler/redis-standalone
---> Fetched chart into workspace /Users/matt/.helm/workspace/charts/redis-standalone
---> Done
```

technosophos:
@adamreese what's the use case for a general settings file? I could probably expand the Repofile to that, but I don't know what else I'd put in there.

I wasn't really intending for a user to edit the Repofile by hand unless to recover from something catastrophic, so I wasn't really intending it to be used like a general settings file. But I guess I could see it as sorta playing the same role as a `.git/config` file. Is that along the lines of what you were thinking?

adamreese:
Yes, It could be more maintainable on the dev side when feature x, y, and z are out and all need config files.

Just my 2 cents...

technosophos:
So what if we did it like this:

Inside of `$HELM_HOME`, add a `config.yaml` file. The structure would look like this:

```
repos:
  default: charts
  tables:
    - name: charts
      repo: https://github.com/deis/charts
```

That would leave the top-level namespace open for other settings alongside `repos`.

Would that be good, @adamreese ?

adamreese:
@technosophos superb

technosophos:
@adamreese Great idea. I'll revise this branch to do that. No point in merging this then having to break things again.

technosophos:
Closing this and pushing in a new version on a separate PR

PR #168, opened 2015-11-04
add linters to test-style target

mboersma:
Adds a `make test-style` target similar to deis/deis. Runs `gofmt`, `golint`, and `go vet`. See also #113.

TODO:
- [x] Should this also be run by default with `make test`? *Edit:* survey so far says "yes"
- [x] Does this need to vendor in `go vet` and `golint` with `glide` or `go get` or...um... *Edit:* no, just `golint` in Travis CI
- [x] `make test` fails for me already in master, see below *Edit:* fixed by wiping out and starting over with $GOPATH + `glide`
- [x] Fix travis CI error about `read` binary

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ install:
   - make bootstrap build
   - popd
   - export PATH="$PATH:$GLIDE_DIR"
+  - go get github.com/golang/lint/golint
 
 script: make bootstrap build test test-charts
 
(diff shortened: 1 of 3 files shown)


technosophos:
`glide -y glide-full.yaml up` to fix your failed build.

We need to amend the Travis setup to use `golint`.

And I'd be fine with this being a prereq for `make test`. I was thinking that we'd use `make quicktest` for quick dev testing, and `make test` for more complete test runs.

mboersma:
Looks like somehow the `read` call fails on Travis CI when doing our `gofmt` check. I'll try to fix that.

mboersma:
I fixed the `gofmt` check not to use `read` so it works on Travis (probably bash vs. sh subshell issue). Tested with a misformatted go file commit, then dropped that commit and force-pushed. Ready for review again.

PR #170, opened 2015-11-04
fix(glide): add deis/pkg to glide.yaml

bacongobbler:
this was quite simple to do. `glide up` now works, but we need to get `make bootstrap` working again as well as documenting how to re-generate glide-full.yaml so other devs know how to do this in the future. Ping @technosophos

### DIFF
--- a/glide-full.yaml
+++ b/glide-full.yaml
@@ -1,29 +1,87 @@
 parent: null
 package: github.com/deis/helm
 import:
-- package: github.com/mitchellh/mapstructure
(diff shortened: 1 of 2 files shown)


technosophos:
`glide up -f glide-full.yaml` will update and then write the resulting data into the file.

bacongobbler:
all set.

technosophos:
The check failed because of a rate limit issue. Is there a way to re-submit the job?

bacongobbler:
Yep! I just logged into travis and hit the "restart build" button.

PR #171, opened 2015-11-04
feat(*): add repo commands and Helm config

technosophos:
This adds the following:

- `helm repo add|list|remove`
- A `$HELM_HOME/config.yaml` file
- Support for `fetch`/`install` out of an alternate repo


bacongobbler:
awesome! I'll play around with this and see what kind of docs I can come around with this, as I'd imagine there may be some improvements to the chart developer workflow with this PR.

mboersma:
I reviewed the code and tested this according to all the bullet points described above--seems to work. :anchor:

action/helper_test.go
```diff
@@ -6,11 +6,23 @@ import (
...
+  tables:
+    - name: charts
+      repo: https://github.com/deis/charts
+`
+
 var helmRoot string
 
 func init() {
+	// Turn on debug output, convert os.Exit(1) to panic()
+	log.IsDebugging = true
```
adamreese: is there a way to set this with the verbose flag `go test -v`?

technosophos: Actually, this is done for an interesting reason: if IsDebugging is true, instead of calling `os.Exit(1)`, `log.Die` calls `panic(err)` and gives a nice pretty stack trace

adamreese: wicked

PR #173, opened 2015-11-04
fix(release): skip update version checks on Travis

bacongobbler:
this is done to prevent rate limiting by github until we find a better
solution.

Sorry @technosophos!


technosophos:
@bacongobbler can you check #175? I just mocked enough of the GH API service that we don't have to call out to it. The test is sorta worthless, but... it avoids this problem.

bacongobbler:
closed in favour of #175

Issue #174, opened 2015-11-04
Add flag to skip Helm version check


technosophos:
On `helm update`, add a `--no-version-check` flag to disable Helm version checks.

This will avoid errors when a helm client cannot talk to Github's API server.

PR #175, opened 2015-11-04
fix(*): add no-version-check flag to update

technosophos:
Addresses #174

### DIFF
--- a/action/update_test.go
+++ b/action/update_test.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/deis/helm/log"
 	"github.com/deis/helm/release"
+	"github.com/google/go-github/github"
 )
 
 func TestEnsurePrereqs(t *testing.T) {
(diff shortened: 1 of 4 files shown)


adamreese:
```
% make test                                                                                                                                                  
go test -v ./. ./manifest ./action ./log ./chart ./dependency ./release
../../google/go-github/github/github.go:22:2: cannot find package "github.com/google/go-querystring/query" in any of:
	/usr/local/Cellar/go/1.5.1/libexec/src/github.com/google/go-querystring/query (from $GOROOT)
	/Users/adamreese/p/go/src/github.com/google/go-querystring/query (from $GOPATH)
make: *** [test] Error 1

% make bootstrap                                                                                                                                             [f77750f] ?? ruby-2.2.3
glide -y glide-full.yaml up
Oops! interface conversion: yaml.Node is yaml.Map, not yaml.List
make: *** [bootstrap] Error 1
```

technosophos:
Upgrade Glide to 0.7

adamreese:
rad

bacongobbler:
my only concern is that we're relying on github to ensure that the v3 API never changes, but they've been great with that so I'm okay with mocking out their API.

PR #178, opened 2015-11-04
style(doc) align readme and architecture

slack:
Align README and architecture with lightly modified cli output.


bacongobbler:
You forgot your colon! ;)

Could you please add one right after the (doc) like this:

`style(doc): align readme and architecture`

Thanks!

PR #186, opened 2015-11-05
feat(*): add 'home' subcommand

technosophos:
This is for shell scripts that need to determine the location of
Helm's home directory.

Example: 

```
cd $(helm home)
```

### DIFF
--- a/helm.go
+++ b/helm.go
@@ -262,6 +262,14 @@ list all available charts.
 			},
 		},
 		{
+			Name:      "home",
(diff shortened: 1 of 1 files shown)


technosophos:
@sgoings would you mind just sanity checking this idea? I am trying to make it easier to script against Helm. Is this a good way of doing that?

mboersma:
Seems potentially useful, and works for me:
```console
$ helm home
/Users/matt/.helm
```

michelleN:
+1

PR #187, opened 2015-11-05
[WIP] feat(Makefile): use pretty version identifiers

### LINKED ISSUE #141: Makefile: Pretty version
Ideally, what would be nice is to have the `VERSION` variable in the Makefile be set according to the following rules:

IF the current commit ID has an associated tag, VERSION should be set to that tag: `1.0.1`

IF the current version ID does not have a tag, VERSION should be set to the last tag plus a "build id" with the short commit: `1.0.1+ee45c49f`

Since the build ID is not considered significant by SemVer 2, users will never be prompted to update based on Git commit alone. This also allows us to use `-alpha` and `-beta` in tags, while still remaining SemVer 2 compliant.

And most importantly, official releases will look pretty `version: 1.0.1`, while dev environments will have build info conveniently embedded in the version.

mboersma:
Simplifies the `VERSION` var in the Makefile to only include the `git` SHA if the commit isn't the same as a tag. Easy thanks to [`git describe`](https://git-scm.com/docs/git-describe). 

This assumes there is at least one tag already in the repository, and the format differs slightly from what was requested, since it includes the # of commits since the tag before the short SHA.

Closes #141.

```console
$ make build
go build -o bin/helm -ldflags "-X main.version=0.1.0-7-g7f8549d" helm.go
$ bin/helm --version
helm version 0.1.0-7-g7f8549d
$ git checkout 0.1.0
$ # (backport this change into the Makefile)
$ make build
go build -o bin/helm -ldflags "-X main.version=0.1.0" helm.go
$ bin/helm --version
helm version 0.1.0
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,6 @@
-VERSION := $(shell git describe --tags --abbrev=0 2>/dev/null)+$(shell git rev-parse --short HEAD)
+VERSION := $(shell git describe --tags 2>/dev/null)
 DIST_DIRS := find * -type d -exec
 export GO15VENDOREXPERIMENT=1
-
-ifndef VERSION
-  VERSION := git-$(shell git rev-parse --short HEAD)
-endif
 
 bootstrap:
 	glide -y glide-full.yaml up


bacongobbler:
looks exactly like what we wanted. :+1:

bacongobbler:
needs a rebase though :)

mboersma:
This is actually somewhat deficient WRT the goals of #141, I realize after discussing with @technosophos. Having a dash-then-number can be confusing to semver, which expects something more like "0.1.0-dev+123456". And really once 0.1.0 is tagged, we'd like VERSION to start with 0.1.**1**-dev. Doing this bit of semver math in Makefile shell is probably feasible but will require some more thought.

sgoings:
Looks similar to what we had before: :smile: https://github.com/deis/helm/commit/32686d2ee35339287edbedfda73304f4a9ac5879

mboersma:
I may not have time to get back to this, so I'm closing it.

PR #201, opened 2015-11-08
Readme update with bintray URLs pointing to .../view#files

rimusz:


sgoings:
I'm not sure this behavior gives us what we want... This will send people to the list of all of the versions we have produced. What about just adding #files after _latestVersion... Will that work?

rimusz:
@sgoings that way works too, shall I rebase my PR?

bacongobbler:
If you can modify your commit messages such that they follow the style guidelines, that would be perfect :)

rimusz:
closing in favour of #203

PR #204, opened 2015-11-09
[WIP] feat(plugin): add plugin support

technosophos:
Following the CLI style of Git, this provides a simple extension mechanism that allows you to put stand-alone tools on `$PATH` and then have them execute "as if" they were part of Helm.

Example:

```
$ helm foo
```

Behind the scenes, this searches for `$PATH/helm-foo`. It passes that command a number of environment variables for ease of use.

Use case: Make it easy for teams to integrate custom tooling into Helm.

### DIFF
--- a/action/plugin.go
+++ b/action/plugin.go
@@ -1,0 +1,52 @@
+package action
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	"github.com/deis/helm/log"
+)
+
+// Plugin attepmts to execute a plugin.
+//
+// It looks on the path for an executable named `helm-COMMAND`, and executes
+// that, passing it all of the arguments received after the subcommand.
+//
+// Output is passed directly back to the user.
+//
+// This ensures that the following environment variables are set:
+//
+//	- $HELM_HOME: points to the user's Helm home directory.
+// 	- $HELM_DEFAULT_REPO: the local name of the default repository.
+func Plugin(homedir, cmd string, args []string) {
+	if abs, err := filepath.Abs(homedir); err == nil {
+		homedir = abs
+	}
+	os.Setenv("HELM_HOME", homedir)
+	os.Setenv("HELM_DEFAULT_REPO", mustConfig(homedir).Repos.Default)
+
+	cmd = PluginName(cmd)
+	execPlugin(cmd, args)
+}
+
+func HasPlugin(name string) bool {
+	name = PluginName(name)
+	_, err := exec.LookPath(name)
+	return err == nil
+}
+
+func PluginName(name string) string {
+	return "helm-" + name
+}
+
+func execPlugin(name string, args []string) {
+	cmd := exec.Command(name, args...)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	cmd.Stdin = os.Stdin
+
+	if err := cmd.Run(); err != nil {
+		log.Die(err.Error())
+	}
+}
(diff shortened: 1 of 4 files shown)


sgoings:
I'm all for extensibility... but is this a feature that starts to get `helm` involved with things that shouldn't be under its concern? i.e. with plugins `helm` starts to look like a lifecycle management tool for applications on kubernetes... do we really want to face these challenges with a package manager?

technosophos:
You raise a good point. If we allow people to write plugins, they may write things that we think are outside of the scope of Helm. But I think that allowing plugins provides three benefits:

- It makes it possible for us to add experimental features as separate plugins, and see whether they're used.
- It will allow users to add internal features specific to their own needs.
- It will allow others to provide extensions to Helm without requiring our buy-in. This would allow some Helm users, for example, to try out a template system.

My gut feeling is that these benefits are worth it, even though I think what you say is true. And I feel like as long as we maintain our own goal of keeping Helm a package manager, we are still staying true to our objective.

rimusz:
@technosophos I totally agree with your explained benefits of having plugins in helm.

bacongobbler:
I'm on board. dash-separated-commands are an easy way for users to plug into their favourite software in their languae of choice. `deis` abuses this and has seen some success with plugins, so I'm in favour of this approach. :hammer:

technosophos:
Added a `helm-sec` command to test out the plugin theory while also building what I hope is a useful "extension" for creating and working with charts. It eases the generation of secrets files.

technosophos:
After a long talk with @sgoings on Slack, I think we should go ahead with this version of plugins, but clearly bill it as experimental.

action/plugin.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	"github.com/deis/helm/log"
+)
+
+// Plugin attepmts to execute a plugin.
```
bacongobbler: attempts*

plugins/sec/secret.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Secret
```
bacongobbler: is this an oopsie generated by `helm sec`? :)

technosophos:
Fixed @bacongobbler 's suggestions

bacongobbler:
accidentally added a binary file (`plugins/sec/helm-sec`). I'm guessing this was a build of the plugin?

technosophos:
just removed that, too.

action/plugin.go
```diff
@@ -0,0 +1,52 @@
...
+		homedir = abs
+	}
+	os.Setenv("HELM_HOME", homedir)
+	os.Setenv("HELM_DEFAULT_REPO", mustConfig(homedir).Repos.Default)
+
+	cmd = PluginName(cmd)
+	execPlugin(cmd, args)
+}
+
+func HasPlugin(name string) bool {
```
mboersma: This should have a documentation comment since it's an exported func.

action/plugin.go
```diff
@@ -0,0 +1,52 @@
...
+	name = PluginName(name)
+	_, err := exec.LookPath(name)
+	return err == nil
+}
+
+func PluginName(name string) string {
```
mboersma: This should have a documentation comment since it's an exported func.

mboersma:
This PR should add "plugins" to the `GO_PACKAGES` in the Makefile so `make test-style` and friends run over that code as well.

technosophos:
/me continues to unquestioningly add every feature @mboersma suggests.

PR #219, opened 2015-11-12
ref(.travis.yml): use glide binary releases

mboersma:
It shouldn't be necessary to build [`glide`](https://github.com/Masterminds/glide/releases) from source each time we run tests. This has Travis CI download and untar the binary as is done in [deis/workflow](https://github.com/deis/workflow).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 env:
-  - GLIDE_DIR="/home/travis/gopath/src/github.com/Masterminds/glide" GO15VENDOREXPERIMENT=1
+  - GO15VENDOREXPERIMENT=1
 
 branches:
   only:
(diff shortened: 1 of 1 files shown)


sgoings:
Mucho bueno! As a sidenote... I'm very interested in trying to get [as much setup of dependencies out of the .travis.yml file](https://www.thoughtworks.com/radar/techniques/programming-in-your-ci-cd-tool) and into the Makefile proper... as it gives us the best ability to move across CI systems, replicate failure locally, etc. etc.

Ideally we should have something even nicer than the [Kubernetes .travis.yml](https://github.com/kubernetes/kubernetes/blob/master/.travis.yml), but we'll get there. Not necessary to improve with this PR, just thoughts from a wandering nerd mind.

mboersma:
Also note that I wanted to use `sudo` to install `glide` to /usr/local/bin, but that would cause Travis CI to kick us off the containerized testing infrastructure onto the "legacy" builders, so I balked.

sgoings:
Yeah I've balked in the same way. I'm not sure if I like this idea, but https://travis-ci.org/deis/helm#L95 shows that you have a user-writeable path in in `PATH` already. :wink:

Issue #213, opened 2015-11-11
Dependencies should have a `repo` attribute


technosophos:
In the `dependencies:` section, there should be an optional `repo:` attribute that can be used to reference a chart's repository (See @olalonde's suggestion at the beginning of #212). When resolving charts, the rule should be:

- If no repo is named, the repo of the referencing chart is used
- If a repo is named, Helm should check for that repo locally...
   - If found, Helm should resolve the name/version within that repo
   - If not found, Helm should emit a warning to the user and exit

The revised structure of the `dependencies:` YAML would look like this:

```yaml
dependencies:
  - name: foo
    repo: github.com/foo/bar
    version: ~1.2.3
```

We should discuss whether we want to add the ability to append a treeish (commit, branch, tag) to the `repo` URI.

Issue #212, opened 2015-11-11
PROPOSAL: helm init / helm install --save


olalonde:
Disclosure: I just discovered Helm and don't really know what I'm talking about except that Helm looks awesome. I've been primarily developing in Node.js for the past few years and npm is a big reason for that (so I'm a bit biased for its way of doing things). Just thought I'd share my perspective as a npm user. 

## unofficial dependencies

https://github.com/deis/charts is nice as an official repository since it leverages existing infrastructure (github) and tools (git). It would be nice in the future to be able to reference dependencies by `githubuser/githubrepo` / absolute URLs to support private/unofficial charts. e.g:

```yaml
dependencies:
  - name: nginx
    version: olalonde/nginx#commit-ish (tag, commit hash, etc.)
```

or 

```yaml
dependencies:
  - name: nginx
    version: git+ssh://user@hostname:project.git#commit-ish
```

I assume this is already planned/implemented, but it would be nice if the [semver v2](http://semver.org/) standard was followed for versioning. npm supports an extensive syntax for specifying [version ranges](https://docs.npmjs.com/misc/semver).

## helm init

npm has a useful command that setups up the initial package.json using information from the current environment (git user / remote origin, current directory name, etc.). Similarly, it would be nice to have a 
`helm init` to initialise the `Chart.yml` file when you want to package a service for helm.

## helm install --save

If I understand correctly, `helm install` installs a chart on a remote kubernetes cluster and caches the chart  in `~/.helm/workspace`? 

What if entire systems could be modelled as charts that live under version control, can be re-installed from scratch with a simple `helm install .` and could be attached as dependencies in larger systems?

The idea would be that every `helm install --save` (in addition to the default behaviour) would add the installed chart to the local `Chart.yml`'s dependencies key. That Chart.yml would essentially contain all the information necessary to re-install your cluster from scratch and it could live under version control (which in turn means history and better collaboration).

Just food for thought...

technosophos:
As you probably noticed, I created a feature request based on your first suggestion.

I'm thinking hard about this idea for creating an analog to `npm install`. It's a slightly different design than what we originally had in mind, but maybe it's a good direction to go.

olalonde:
Thanks, those were just my first impressions. A distributed package manager is obviously going to be very different from a programming language one.

mboersma:
The future of Helm is at [kubernetes/helm](https://github.com/kubernetes/helm). Helm Classic is in maintenance mode, so I'm closing this proposal.

Issue #195, opened 2015-11-06
Proposal: Extended dependency declarations


technosophos:
Currently we have dependency declarations like this:

```
depends:
  - name: redis    # chart name
    version: ~1.3  # semver version or range
```

We need to extend the dependency system to support `provider` support. (For info on providers, see https://github.com/deis/helm/blob/master/docs/using_labels.md)

One method of doing this is to extend the `depends` structure to have the following fields:

- `type`: One of `provider` or `chart`, with `chart` being the default.
- `name`: The name. Required by `type:chart`, optional on `type:provider`
- `version`: A SemVer range, applicable only to `type: chart`
- `selector`: A label selector (http://kubernetes.io/v1.0/docs/user-guide/labels.html), used only by `type: provider`

bacongobbler:
Given the following use case with this proposal:

 - I have a chart named foo
 - foo requires a mysql chart
 - in my charts I have `mysql` and `mysql-cluster`, both are mysql providers

How would `helm install foo` work when there's a provider collision?

> type: One of provider or chart, with chart being the default.

Is there a use case when we would want our chart to be dependent on another chart?

mboersma:
The future of Helm is at [kubernetes/helm](https://github.com/kubernetes/helm). Helm Classic is in maintenance mode, so I'm closing this proposal. 🚪

Issue #194, opened 2015-11-06
Annotations with Helm metadata


technosophos:
The following Helm data should definitely be put into the annotations field on manifests:

- version
- description

Possibly we should put the entire Chart.yaml into the annotation.

technosophos:
This is to be done on install (?).

technosophos:
I am officially pulling this off of the 0.2 milestone.

technosophos:
Also add Chart name.

NB: It looks like the current recommendation for namespacing Annotations is to use a domain/path structure:

```
helm.sh/foo/bar: value
```

PR #220, opened 2015-11-13
feat(*): add YAML/JSON codec

technosophos:
The codec provides a faster and easier way of encoding and decoding
both YAML and JSON.

### DIFF
--- a/action/install.go
+++ b/action/install.go
@@ -1,15 +1,16 @@
 package action
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
 
 	"github.com/deis/helm/chart"
+	"github.com/deis/helm/codec"
 	"github.com/deis/helm/dependency"
 	"github.com/deis/helm/log"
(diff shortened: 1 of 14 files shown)


technosophos:
The idea with this PR is to (a) improve our YAML encoder, while (b) reducing our usage of an assortment of Kubernetes APIs. It still uses the type system, but it uses a more straightforward YAML/JSON codec. It borrows a little of the Kubelite code, but without replacing the core Kubernetes API.

This commit is also prep for the new types added in k8s v1.1.1, and for #194, #195, #212, and #213 (all of which require better YAML encoding).

action/install.go
```diff
@@ -1,15 +1,16 @@
...
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
 
 	"github.com/deis/helm/chart"
+	"github.com/deis/helm/codec"
```
mboersma: This package URL needs to be updated to helm/helm. :anchor:

PR #224, opened 2015-11-15
ref(glide) move glide-full.yaml to glide.yaml

michelleN:
It didn't look like we were using the old glide.yaml file anymore so I replaced the glide.yaml file with glide-full.yaml, renamed it to glide.yaml, and made the necessary change in the make bootstrap command. Would love any feedback.


mattfarina:
I'll wait for @technosophos to weight in here. I think what was going on was `glide-full.yaml` was being used as a lock file generated from `glide.yaml`. So, to update the dependencies `glide up` was being run. Then for future installs of that pinned version `glide -y glide-full.yaml up` was being used.

I'm just starting on handling a lockfile to automate this and this was a hack around until that as in place. If you want to find me in the IRC room #masterminds or #deis I'm happy to talk about this in more detail.

technosophos:
Yeah, @mattfarina is right about my intentions here. I was using `glide-full` as a lock file until Glide 0.8 fixed this in the way I think is the right way.

We could do what @michelleN suggests, but I would kinda rather hang back until the Glide lock file thing gets settled. That should solve the deeply nested dependencies problem while still keeping our top glide file accurate about exactly which things Helm depends directly on.

michelleN:
@technosophos @mattfarina alrighty, sounds good. I'm fine with hanging back as well. I'll close this. Thanks!

PR #225, opened 2015-11-15
ref(update): remove depricated ensureRepo func

michelleN:


mboersma:
There is still an [`ensureRepo`](https://github.com/deis/helm/blob/master/config/config.go#L172) func in config/config.go that's nearly identical and is being called. Should we take the test from update_test.go and add it to config_test.go instead of deleting it?

michelleN:
@mboersma yeah that's a good call. I'll refactor.

config/config_test.go
```diff
@@ -1,10 +1,25 @@
...
+func createTmpHome() string {
+	tmpHome, _ := ioutil.TempDir("", "helm_home")
+	defer os.Remove(tmpHome)
+	return tmpHome
+}
+
+func TestEnsureRepo(t *testing.T) {
+	tmpHome := createTmpHome()
+
+	repo := "https://github.com/deis/charts"
```
bacongobbler: now we're located at helm/charts :)

michelleN: alrighty! thanks.

michelleN:
rebased and ready :)

Issue #237, opened 2015-11-17
No support for kind ServiceAccount


jchauncey:
```
[WARN] No support for kind ServiceAccount. Ignoring.
---> Running `kubectl create -f` ...
```

The manifest is taken from here - https://github.com/pires/kubernetes-elasticsearch-cluster/blob/master/service-account.yaml

technosophos:
I'll fix that.

PR #244, opened 2015-11-17
fix(*): add support for ServiceAccount

technosophos:
This addresses #237.


bacongobbler:
looks like an accidental commit got squished in here. Would you mind squashing it?

Issue #217, opened 2015-11-12
helm remove should be able to remove multiple charts


sgoings:
Currently `helm remove` only can support the removal of one chart from the working directory.

technosophos:
And as related to #177, it should prompt just once to uninstall all of them, right?

PR #245, opened 2015-11-17
feat(rm): allow more than one chart removal

technosophos:
This addresses #217.


mboersma:
This should also add a multiple-chart test to remove_test.go.

Issue #183, opened 2015-11-05
Check if chart is installed before helm remove


adamreese:
If a user removes a chart from their working directory, should helm check to see if is installed?

bacongobbler:
I accidentally removed a chart and couldn't remove the pod running on kubernetes using helm. I had to `k delete po redmine`. I think prompting before removing would also be a safe counter-measure before I blindy fat-finger `helm remove` instead of `helm uninstall`.

PR #257, opened 2015-11-18
fix(remove): prevent removing installed charts from workspace

adamreese:
Fixes: https://github.com/helm/helm/issues/183

Introduces a `--force` option on `helm remove` to remove charts from workspace that are installed.

TODO
- [x] handle kubectl not connected to cluster

### DIFF
--- a/action/remove.go
+++ b/action/remove.go
@@ -1,25 +1,88 @@
 package action
 
 import (
+	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"strings"
 
 	"github.com/helm/helm/log"
+	"github.com/helm/helm/manifest"
 )
+
+// kubeGetter wraps the kubectl command, override in tests
+type kubeGetter func(string) string
+
+var kubeGet kubeGetter = func(m string) string {
+	log.Debug("Getting manifests from %s", m)
+
+	a := []string{"get", "-f", m}
+	out, err := exec.Command("kubectl", a...).CombinedOutput()
+	if err != nil {
+		log.Err("Could not get manifest. %s", err)
+	}
+	return string(out)
+}
 
 // Remove removes a chart from the workdir.
 //
 // - chart is the source
 // - homedir is the home directory for the user
-func Remove(chart string, homedir string) {
+// - force will remove installed charts from workspace
+func Remove(chart, homedir string, force bool) {
 	chartPath := filepath.Join(homedir, WorkspaceChartPath, chart)
 	if _, err := os.Stat(chartPath); err != nil {
 		log.Die("Chart not found. %s", err)
 	}
 
-	if err := os.RemoveAll(chartPath); err != nil {
-		log.Die("%s", err)
+	var connectionFailure bool
+
+	// check if any chart manifests are installed
+	installed, err := checkManifests(chartPath)
+	if err != nil {
+		if strings.Contains(err.Error(), "unable to connect") {
+			connectionFailure = true
+		} else {
+			log.Die(err.Error())
+		}
 	}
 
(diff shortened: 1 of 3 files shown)


action/remove.go
```diff
@@ -2,24 +2,70 @@ package action
...
+	"strings"
 
 	"github.com/helm/helm/log"
+	"github.com/helm/helm/manifest"
 )
 
+// kubeGet executes the kubectl command, override in tests
+var kubeGet func(string) string
+
+func init() {
```
technosophos: Do you need an init block for this, or could you simply do the `= func...` part on line 14?

adamreese: I'll give it a shot

action/remove.go
```diff
@@ -2,24 +2,70 @@ package action
...
+	"os/exec"
 	"path/filepath"
+	"strings"
 
 	"github.com/helm/helm/log"
+	"github.com/helm/helm/manifest"
 )
 
+// kubeGet executes the kubectl command, override in tests
+var kubeGet func(string) string
```
technosophos: The most idiomatic way to do this would be to add a type:

```go
type kubeGetter func(string) string
var kubeGet kubeGetter
```

A good example in `"net/http".HandlerFunc`.

action/remove.go
```diff
@@ -2,24 +2,70 @@ package action
...
+
+func init() {
+	kubeGet = func(m string) string {
+		log.Debug("Getting manifests from %s", m)
+
+		a := []string{"get", "-f", m}
+		out, _ := exec.Command("kubectl", a...).CombinedOutput()
```
technosophos: Why not catch and log the error on the exec?

adamreese: Because I forgot to add it back in .__.

action/remove.go
arschles: +1 to adding the type.

I'd like to propose an alternative to the global implementation if I may. It seems that `kubeGet` is only used once in this file. I didn't check if it's used elsewhere but my proposal would stand for any other locations in code as well.

Instead using a global `kubeGetter`, my alternative is passing the `kubeGetter` into the `func`s that require it, and then passing the global implementation as needed from production code. Test code can then pass test implementations as needed.

For example, the global `func` is used [here](https://github.com/helm/helm/pull/257/files#diff-ac2075f1a23904bf5eb1df751e18eb6cR63). My suggesting is to add a `kubeGetter` parameter to the [`checkManifests`](https://github.com/helm/helm/pull/257/files#diff-ac2075f1a23904bf5eb1df751e18eb6cR53) func and having code pass the implementation manually, and have test code pass a test version.

This proposed change will make the code more explicit in what it's doing.

technosophos: I could see doing things that way. I guess my main objection to that would be that we don't really want layers _above_ Remove to be required to know the internals of how Remove works. So I don't think passing the function into Remove would be a good solution.

My read of what @adamreese is doing here is just making it possible to override the external call for the purposes of unit testing. This does feel a little elaborate if that is the only actual use case... but I don't have a better solution.

I'd still reference the way the HTTP library handles this as generally acceptable for setting a default handler for something.

arschles: fair point. I may be sliding too far into pedantry at this point, global way is ok with me too. I'm still +1 on the custom type, just as `net/http` does as you said.

adamreese: This was a temporary fix for not being able to mock kubectl responses.  My goal is to have a pattern that can be used in other actions as well.

arschles: @adamreese I was told by @technosophos that you were working on pulling out kubectl functionality into an interface, which would provide a good way to mock as you said. do you have that work somewhere in a PR?

Issue #177, opened 2015-11-04
Uninstall should prompt for confirmation (and have a force flag)


technosophos:
Because `helm uninstall` performs an operation on the k8s cluster, we should really prompt for confirmation before deleting. I'm envisioning something like this:

```
helm uninstall alpine
---> Components to be uninstalled:
  pod/alpine
  service/alpine-gateway
Are you sure you want to delete these 2 items (Y,n): y
---> Running `kubectl delete` ...
```

To sidestep, the user can provide a `--force` flag

bacongobbler:
+1

michelleN:
@technosophos Could we go a little further to ask "This will run a `kubectl delete` operation. Are you sure you want to delete these 2 items (Y,n):"

sgoings:
@michelleN we could do that, or follow the `rm -i` method which is a prompt for every file (in our case: chart) removal.

michelleN:
:+1:

technosophos:
The difference between `rm -i` and something like this is that the units you remove with `rm` are not necessarily interdependent.

For charts, when you deploy a chart, you assume coupling between all of these things. So can we accurately do a thing-by-thing deletion and assure the user that we won't leave them in a broken state?

I feel like if the answer is no, we should not allow it. They can always selectively delete using kubectl itself.

arschles:
my $0.2 is that if we understand the dependency graph and we do chart by chart deletion, then we should delete in dependency order. that seems like a lot of logic (right now) just to do something that users could do themselves with `kubectl`.

maybe instead it's a part of a larger issue to add more dependency related commands to helm?

arschles:
unrelated request - it would be nice to add a `-y` flag to `helm uninstall` so that you can run it in CI systems

technosophos:
The issue here is that a specific chart may be deployed into Kubernetes, and have several associated manifests (a service, a few RCs, a volume, some secrets...). If we delete, say, a service but not the RCs, we could leave things in an uncertain state. Or if we keep an RC and delete the volume and secret it uses, we would at best get cluster errors, and at worst get really weird side-effects.

arschles:
I see, I was thinking in terms of service => service, or even chart => chart dependencies

PR #260, opened 2015-11-19
fix(uninstall): prompt for confirmation

technosophos:
Addresses #177

### DIFF
--- a/action/uninstall.go
+++ b/action/uninstall.go
@@ -23,7 +23,7 @@ import (
 // 	- Volumes
 // 	- Secrets
 //	- Namespaces
-func Uninstall(chartName, home, namespace string) {
+func Uninstall(chartName, home, namespace string, force bool) {
 	if !chartFetched(chartName, home) {
 		log.Info("No chart named %q in your workspace. Nothing to delete.", chartName)
 		return
@@ -37,7 +37,7 @@ func Uninstall(chartName, home, namespace string) {
 	if err := deleteChart(c, namespace, true); err != nil {
 		log.Die("Failed to list charts: %s", err)
 	}
-	if !promptConfirm("Uninstall the listed objects?") {
+	if !force && !promptConfirm("Uninstall the listed objects?") {
 		log.Info("Aborted uninstall")
 		return
 	}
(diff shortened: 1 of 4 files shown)


action/uninstall.go
```diff
@@ -30,50 +34,126 @@ func Uninstall(chartName, home, namespace string) {
...
+	return x.w.Write(b)
+}
+
+func deleteChart(c *chart.Chart, ns string, dry bool) error {
 	// We delete charts in the ALMOST reverse order that we created them. We
 	// start with services to effectively shut down traffic.
 	ktype := "service"
 	for _, o := range c.Services {
-		if err := kubectlDelete(o.Name, ktype, ns); err != nil {
+		if dry {
```
sgoings: Is there a better way to handle dry run logic in Go? It seems like it could be abstracted cleanly in some manner instead of sprinkling if statements around...

technosophos: I'm going to work on this for a separate PR. There are several cases of that same pattern that I think should all be fixed. But I didn't want to broaden the scope of this issue too far.

arschles: @adamreese is working on abstracting the `kubectl` commands behind an interface, so once that's done it would be pretty easy to pass around the interface and pass a 'dry run' implementation everywhere if `dry` is set

action/uninstall.go
```diff
@@ -30,50 +34,126 @@ func Uninstall(chartName, home, namespace string) {
...
+func promptConfirm(msg string) bool {
+	oldState, err := terminal.MakeRaw(0)
+	if err != nil {
+		log.Err("Could not get terminal: %s", err)
+		return false
+	}
+	defer terminal.Restore(0, oldState)
+
+	f := readerWriter(log.Stdin, log.Stdout)
+	t := terminal.NewTerminal(f, msg+" (y/N) >> ")
```
mboersma: I found the chevrons `>>` unexpected:
```console:
$ helm uninstall alpine
pod/alpine
Uninstall the listed objects? (y/N) >> n
---> Aborted uninstall
```

Wouldn't it be more conventional to omit them and just leave a space after `(y/N)`? But I don't care that much about it and this PR LGTM otherwise.

technosophos: I agree. I'll remove them.

technosophos:
This is ready for review again.

log/log.go
```diff
@@ -15,6 +15,11 @@ var Stdout io.Writer = os.Stdout
 // Stderr is the logging destination for error messages.
 var Stderr io.Writer = os.Stderr
 
+// Stdinn is the input alternative for logging.
```
sgoings: "Stdinn" -> "Stdin"

technosophos:
@sgoings fixed

Issue #259, opened 2015-11-18
Remove `-p` from `helm install`


technosophos:
The `-p` flag is sorta outside of the normal workflow, and is known to cause problems with dependencies. It currently has a big batch of branching logic to handle the case right now, and it's easy to screw things up this way.

Is there any good reason to keep it around?

@gabrtv @arschles and I discussed, and I feel like we could comfortably get rid of this.

gabrtv:
+1 on removing this.  The `AltInstall` stuff is a :snowflake: and not the pretty kind.  I also think that this will help drive resolution on #249, which has similar motivations.

arschles:
:+1:

technosophos:
Changed from proposal to enhancement.

PR #261, opened 2015-11-19
fix(install): remove -p flag

technosophos:
Addresses #259.


bacongobbler:
didn't make all that much sense. If a user wants to load a chart from a separate directory, either they need to set $HELM_HOME or move the chart into the cache. +1

PR #264, opened 2015-11-20
ref(*): refactor kubectl to a package

adamreese:


helm.go
```diff
@@ -164,6 +165,10 @@ This will not alter the charts in your workspace.
 			ArgsUsage: "[chart-name...]",
 			Action: func(c *cli.Context) {
 				minArgs(c, 1, "uninstall")
+				// TODO: remove global kubectl.Runner
```
arschles: :+1: on removing it. we talked about passing around some kind of context instead of setting a global here. I'm in favor of doing that, as it will allow for us to write tests that are more explicit as to what's going on. I'd also consider that more in line with the "go way"

adamreese:
@arschles made suggested changes

action/install.go
```diff
@@ -201,44 +202,3 @@ func chartFetched(chartName, home string) bool {
...
-	if err := c.Start(); err != nil {
-		return err
-	}
-
-	log.Debug("File: %s", string(data))
-	in.Write(data)
-	in.Close()
-
-	return c.Wait()
-}
```
arschles: I love me some deleted code :+1:

helm.go
```diff
@@ -164,8 +165,12 @@ This will not alter the charts in your workspace.
 			ArgsUsage: "[chart-name...]",
 			Action: func(c *cli.Context) {
 				minArgs(c, 1, "uninstall")
+				client := kubectl.Client
```
arschles: @adamreese do we need the global `kubectl.Client` anymore? asking because this could also just be `client := kubectl.RealClient{}`. not necessary to change in this PR, just for my info

helm.go
```diff
@@ -263,7 +272,7 @@ list all available charts.
 			Usage:     "Displays information about cluster.",
 			ArgsUsage: "",
 			Action: func(c *cli.Context) {
-				action.Target()
+				action.Target(kubectl.Client)
```
arschles: shouldn't this be `action.Target(client)`?

adamreese: `client` is not defined.  This is passing the package global client

kubectl/kubectl.go
```diff
@@ -0,0 +1,44 @@
...
+	"os/exec"
+	"strings"
+)
+
+// Path is the path of the kubectl binary
+var Path = "kubectl"
+
+// Runner is an interface to wrap kubectl convenience methods
+type Runner interface {
+	ClusterInfo() ([]byte, error)
```
arschles: can you put some godocs for each of these interface functions? if possible, it would be nice to run [`golint`](https://github.com/golang/lint) on the `kubectl` package, as it will tell you the major points for improvement

adamreese: I'll add some comments.  `golint` gives me nada

helm.go
arschles: ah, missed that. I think this needs to check the `dry-run` flag then

kubectl/kubectl.go
arschles: hmm, at some point I'll look into why `golint` gives nothing. thanks for the comments!

bacongobbler:
general Go packaging question: is it better to have each implementation of type `Runner` to be in its own file rather than having each action? i.e. kubectl.go, runner.go, printrunner.go, realrunner.go, testrunner.go and printrunner_test.go

I've seen the type-oriented filesystem layout often but not the action-oriented layout (unless it's a CLI package). Is there a specific advantage between the two?

kubectl/kubectl_test.go
```diff
@@ -0,0 +1,12 @@
+package kubectl
+
+type TestRunner struct {
+	Runner
+
+	out []byte
+	err error
+}
+
+func (r TestRunner) Get(stdin []byte, ns string) ([]byte, error) {
```
bacongobbler: I'm confused why we have a `TestRunner` and a `PrintRunner` type. Aren't they both the same, but one just displays whatever you want it to and one actually displays the `kubectl` output? What's the idea behind `TestRunner`?

arschles:
@bacongobbler I don't think there's really a standard here. FWIW I prefer each implementation to live in its own file whenever possible.

action/install.go
```diff
@@ -55,21 +65,19 @@ func Install(chartName, home, namespace string, force bool, dryRun bool) {
...
 	}
 
-	msg := "Running `kubectl create -f` ..."
-	if dryRun {
-		msg = "Performing a dry run of `kubectl create -f` ..."
-	}
-	log.Info(msg)
-	if err := uploadManifests(c, namespace, dryRun); err != nil {
-		log.Die("Failed to upload manifests: %s", err)
+	//@FIXME this output is confusing with --dry-run
```
arschles: then remove it? not blocking this PR's merge

action/install_test.go
```diff
@@ -0,0 +1,79 @@
...
+func TestInstall(t *testing.T) {
+	var output bytes.Buffer
+	log.Stdout = &output
+	log.Stderr = &output
+
+	// Todo: add tests
+	// - with an invalid chart name
+	// - with failure to check dependencies
+	// - with failure to check dependencies and force option
+	// - with chart in current directly
```
arschles: there are enough todos here to warrant a new GH issue. not blocking this PRs merge

kubectl/kubectl_test.go
arschles: @adamreese ping

arschles:
@adamreese I'm giving this a LGTM. I don't see any blocking issues in this code, but I (and others) still have some yet-to-be-answered questions

kubectl/kubectl_test.go
adamreese: `TestRunner` is for testing, `PrintRunner` is for `--dry-run`

adamreese:
I am breaking up this pr into smaller chunks

Issue #248, opened 2015-11-17
.yml vs .yaml


jchauncey:
So I noticed taht I had a manifest with `.yml` as the extension instead of `.yaml`. It was kind of confusing as to why it was never being installed. It would be helpful if we printed a warning or just accepted that extension as well.

arschles:
+1 on accepting both

technosophos:
Pretty sure there are only a few places where we do things like this:

https://github.com/helm/helm/blob/master/manifest/manifest.go#L38

So it would be really easy to accept either extension.

arschles:
I'll take this one

arschles:
Fixed by #306

PR #266, opened 2015-11-20
feat(manifest.go): adding support for yml extensions in manifests (WIP)

arschles:
Fixes https://github.com/helm/helm/issues/248

This does not include support for `.yml` extensions in charts. @sgoings @michelleN @technosophos (others?) I'd like us to support `Chart.yml` at some point but not in this PR, and IMO we should error if both a `Chart.yml` and a `Chart.yaml` exist.

### DIFF
--- a/chart/chart_test.go
+++ b/chart/chart_test.go
@@ -22,7 +22,7 @@ func TestLoad(t *testing.T) {
 	}
 
 	if len(c.Kind["Pod"]) != 3 {
-		t.Errorf("Expected 3 pods, got %d", len(c.Kind["Pod"]))
+		t.Fatalf("Expected 3 pods, got %d", len(c.Kind["Pod"]))
 	}
 
 	if len(c.Kind["ReplicationController"]) == 0 {
--- a/chart/chartfile_test.go
+++ b/chart/chartfile_test.go
@@ -8,7 +8,7 @@ import (
 func TestRepoName(t *testing.T) {
 	name := RepoName(".")
 	if name == "" {
-		t.Errorf("Expected a git URL.")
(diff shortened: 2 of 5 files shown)


sgoings:
Some thoughts:

- only support Chart.yaml
- `helm` prefers seeing `Chart.yaml` over `Chart.yml` (?) and produces a warning if both exist (and tells the user which one it's using)

Although this seems unnecessarily complex. Why can't we be opinionated about what filename extensions we support?

technosophos:
My feeling is that a "default file" should have only one acceptable name. That eliminates any potential confusion.

For non-default files (like manifests), it makes sense to take the opposite approach: support all the standard variants.

arschles:
@technosophos fine with me also

bacongobbler:
+1 for @technosophos's suggestion

arschles:
@technosophos @bacongobbler @sgoings after a rebase, `make bootstrap build test` fails in travis due to a build error but passes on my machine. anyone seen an issue like this before? I have to move onto some deis work now, but will come back to this later

sgoings:
Yikes. I very much dislike the k8s.io dependency problems. We've had them in the past... I wonder if you should rebase all those commits and try again?

bacongobbler:
yeah looks like the commits being all munged up like it is might have to do with it. Other than that I'm clueless because it does say that it fetched updates for k8s.io/kubernetes during `glide up`.

arschles:
hm, I'm at a loss here. everything has been rebased, and it works on a brand new docker container with an empty GOPATH. since this is such a small change, I'm gonna cherrypick changes here onto a new branch & PR

arschles:
closing in favor of #306

Issue #243, opened 2015-11-17
Rename codec/manifest.go to codec/object.go


technosophos:
The filename is a relic of when the Object type used to be called Manifest.

PR #267, opened 2015-11-20
fix(codec): rename manifest to object

technosophos:
Addresses #243


carmstrong:
I think you mean "rename manifest to object"? Also, needs a rebase.

PR #281, opened 2015-11-25
first spike of helm#280 to support OpenShift Templates

jstrachan:
the templates get expanded on the client side before they are uploaded so they work on any kubernetes environment


jstrachan:
gonna close this and submit a clean PR after a massive rebase due to all the recent changes...

Issue #277, opened 2015-11-24
Fix strange Debug test output.


technosophos:
When I run tests in `action`, I see this:

```
=== RUN   TestListRepos
[DEBUG] [DEBUG] [DEBUG] [DEBUG] ---> Creating /var/folders/p5/5940dcgj6jx_w1w_2c9nrfkr0000gn/T/helm_home147762185/config.yaml
[DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] [DEBUG] --- PASS: TestListRepos (0.01s)
=== RUN   TestSearchByName
```

sgoings:
With that much debugging information, you can't lose!

PR #283, opened 2015-11-25
fix(log): print debug and warn messages to stderr

adamreese:
Fixes: https://github.com/helm/helm/issues/277


technosophos:
Does this get rid of the `[Debug] [Debug] [Debug]` messages? I never did ferret out the reason for a long sequence of empty debug messages.

adamreese:
It does.  The test was capturing `Stdout` which was the message, but the `[DEBUG]` was printing to `Stderr`

technosophos:
Oh! That totally makes sense now.

PR #284, opened 2015-11-25
feat(*): Exit(1) when log.Err is called

mattfarina:
This causes helm to always exit(1) when `log.Err()` is called. Even if `log.Err()` is called and helm continues.

### DIFF
--- a/action/remove.go
+++ b/action/remove.go
@@ -30,7 +30,8 @@ var kubeGet kubeGetter = func(m string) string {
 func Remove(chart, homedir string, force bool) {
 	chartPath := filepath.Join(homedir, WorkspaceChartPath, chart)
 	if _, err := os.Stat(chartPath); err != nil {
-		log.Die("Chart not found. %s", err)
+		log.Err("Chart not found. %s", err)
+		return
 	}
 
 	if !force {
--- a/log/log.go
+++ b/log/log.go
@@ -23,6 +23,9 @@ var Stdin io.Reader = os.Stdin
 // IsDebugging toggles whether or not to enable debug output and behavior.
 var IsDebugging = false
 
+// ErrorState denotes if application is in an error state.
+var ErrorState = false
+
 // New creates a *log.Logger that writes to this source.
 func New() *log.Logger {
 	ll := log.New(Stdout, pretty.Colorize("{{.Yellow}}--->{{.Default}} "), 0)
@@ -55,6 +58,7 @@ func CleanExit(format string, v ...interface{}) {
 func Err(format string, v ...interface{}) {
 	fmt.Fprint(Stderr, pretty.Colorize("{{.Red}}[ERROR]{{.Default}} "))
 	fmt.Fprintf(Stderr, appendNewLine(format), v...)
+	ErrorState = true
 }
 
 // Info prints a green-tinted message.
--- a/helm.go
+++ b/helm.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"os"
 
 	"github.com/codegangsta/cli"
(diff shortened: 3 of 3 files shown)


technosophos:
This is very clever, and I think it is a great idea. @adamreese does this make sense to you?

adamreese:
Makes sense to me

action/remove.go
```diff
@@ -33,7 +33,8 @@ var kubeGet kubeGetter = func(m string) string {
 func Remove(chart, homedir string, force bool) {
 	chartPath := filepath.Join(homedir, WorkspaceChartPath, chart)
 	if _, err := os.Stat(chartPath); err != nil {
-		log.Die("Chart not found. %s", err)
+		log.Err("Chart not found. %s", err)
+		return
```
sgoings: Is this return mainly so go lint doesn't complain?

mattfarina: @sgoings `log.Die` as previously used there. It performs an `os.Exit(1)` ending the applications execution. When I switched to `log.Err` it no longer exited the application. The `return` ends the execution of the `Action` handler.

technosophos:
@mattfarina can you rebase this? I think it should be trivial to do the rebase. Then I'll add LGTM2

mattfarina:
@technosophos rebased.

technosophos:
I'm afraid it needs another rebase, @mattfarina. I didn't get to it in time.

mattfarina:
@technosophos rebased again

log/log.go
```diff
@@ -55,6 +58,7 @@ func CleanExit(format string, v ...interface{}) {
 func Err(format string, v ...interface{}) {
 	fmt.Fprint(Stderr, pretty.Colorize("{{.Red}}[ERROR]{{.Default}} "))
 	fmt.Fprintf(Stderr, appendNewLine(format), v...)
+	ErrorState = true
```
arschles: for posterity can you add a note in the godoc that this func is not concurrency safe?

gerred: Or can we refactor this to not use this sort of state? Perhaps app.After selects between an errChan and a default state, and if it gets a message on the errChan it exits with 1.

gerred: Something like this: https://gist.github.com/gerred/3c57d8ce5629a32e3693 @mattfarina

technosophos: Given the general concurrency state of this entire package, it'd be sorta pedantic to require that this one variable be marked as not concurrency safe. After all... look at Stdout and Stderr.

I'm not opposed at all to the channel-based route. In fact, might there be a way to make it an int channel and pass in a specific error code? cli frustrates me in that it seems to only support returning `1`.

gerred: We could, and not change any of the current logic from that gist. The select will make it grow nicely later. Could eventually do a HelmError type that has both the OS exit code and a more specific message, but I doubt it's needed for this particular addition.

arschles: @technosophos I think it's always prudent to document when something is concurrency unsafe. why not do it? anyway, also not opposed to the channel way.

mattfarina: Before I make changes, the state is not in error until one happens somewhere. Once that happens it change to an error state. Doesn't matter if it happens on one goroutine or another. The error state is global to the application.

What is gained from changing this to be a channel?

gerred: Practically from the stance of the code today? Nothing. It does satisfy the Go best practice of "Don't communicate by sharing memory, share memory by communicating", which might be pedantic. @technosophos mentioned this may be nice in the future too for bubbling up more specific errors/error codes later down the line, but I think that's beyond the scope we're talking about here.

arschles: I was asking for a comment as a defensive measure for future maintainability. Anyway, not a showstopper for this PR.

I may go through portions of the codebase and add concurrency related comments or make concurrency safe in a later PR. There's little concurrency (if any) in this codebase right now that I know of, so I'm not considering it high priority.

Issue #285, opened 2015-11-25
Error removing charts when I'm offline.


technosophos:
I want to remove a chart when I'm "offline" from my Kube cluster. I am getting this error:

```
⇒  helm list
---> 	alpine (alpine-pod 0.0.1) - Simple pod running Alpine Linux.
---> 	bazbar (bazbaz 0.0.1) - Simple pod running Alpine Linux.
---> 	bazbaz (bazbaz 0.0.1) - Simple pod running Alpine Linux.
---> 	cassandra (cassandra 0.2.1) - The Apache Cassandra Project.
---> 	example-todo (example-todo 0.0.6) - Example Todo application backed by Redis
---> 	foobar (foobar 0.0.1) - Simple pod running Alpine Linux.
---> 	installp (alpine 0.0.1) - Simple pod running Alpine Linux.
---> 	memcached (memcached 1.2.24-chart.1) - A simple Memcached cluster
---> 	myriak (riak 0.1.0) - Riak standalone pod.
---> 	redis-cluster (redis-cluster 0.0.5) - Highly available Redis cluster with multiple sentinels and standbys.
☃ @ENG001590.local ~/Code/Go/src/github.com/helm/helm «/Users/mattbutcher/Code» ‹feat/objects-1.1*›
⇒  helm rm bazbaz
[ERROR] Could not get manifest. exit status 1
[ERROR] Found 1 installed manifests for bazbaz.  To remove a chart that has been installed the --force flag must be set.
☃ @ENG001590.local ~/Code/Go/src/github.com/helm/helm «/Users/mattbutcher/Code» ‹feat/objects-1.1*›
⇒  helm rm --force bazbaz
[ERROR] Could not get manifest. exit status 1
---> All clear! You have successfully removed bazbaz from your workspace.
```

I am not exactly sure what the right story is re: error messages, but for sure the `--force` should have exited cleanly on this one. /cc @adamreese

PR #289, opened 2015-11-25
fix(rm): remove missleading error on rm

adamreese:
Fixes: https://github.com/helm/helm/issues/285

### DIFF
--- a/action/remove.go
+++ b/action/remove.go
@@ -18,10 +18,7 @@ var kubeGet kubeGetter = func(m string) string {
 	log.Debug("Getting manifests from %s", m)
 
 	a := []string{"get", "-f", m}
-	out, err := exec.Command("kubectl", a...).CombinedOutput()
-	if err != nil {
-		log.Err("Could not get manifest. %s", err)
-	}
+	out, _ := exec.Command("kubectl", a...).CombinedOutput()
 	return string(out)
 }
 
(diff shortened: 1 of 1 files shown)


action/remove.go
```diff
@@ -18,10 +18,7 @@ var kubeGet kubeGetter = func(m string) string {
 	log.Debug("Getting manifests from %s", m)
 
 	a := []string{"get", "-f", m}
-	out, err := exec.Command("kubectl", a...).CombinedOutput()
-	if err != nil {
-		log.Err("Could not get manifest. %s", err)
-	}
+	out, _ := exec.Command("kubectl", a...).CombinedOutput()
```
arschles: +1 that it was a misleading error, but can we at least still `log.Warn` the error so it's brought to the surface?

also, if there was an error, `out` will be an empty `[]byte` slice. will callers expect that?

technosophos: Well... I guess the real point is that it shouldn't execute that command at all in the case where `--force` is specified.

adamreese: @arschles `out` is not empty on error.  the error that gets returned is `exit status 1` which is not useful for the user.

arschles: ok, fair enough. In general, I don't like seeing us ignoring errors. it would be nice to combine the two outputs on error like so:

```go
out, err := exec.Command("kubectl", a...).CombinedOutput()
if err != nil {
  return fmt.Sprintf("%s (%s)", string(out), err)
}
return string(out)
}
```

Issue #282, opened 2015-11-25
support for annotation based secret generation


jstrachan:
we found when making charts that often they require secrets to be imported or created and sometimes secrets are shared between charts. 

e.g. when using Gerrit, Jenkins and SonarQube we need to share public keys to Gerrit for things like Jenkins and SonarQube which have their own generated/imported public/private keys. Or to install our full CI / CD system so it can do real releases to github, maven central and docker hub we need to import our real SSH & GPG keys.

So we came up with a simple set of annotations we could put onto our charts:
https://github.com/fabric8io/fabric8/blob/master/docs/secretAnnotations.md

which a tool can then process to import (if they are on disk) or generate Secrets if they don't exist as charts are installed. Then users can install/setup the secrets however they like before you install things; otherwise the install just works and worst case you get auto-generated secrets in each environment.

Up to now we've been using the `gofabric8 secrets` command to do this. e.g. here's the code...
https://github.com/fabric8io/gofabric8/blob/master/cmds/secrets.go#L84

we basically iterate through all installed OpenShift Templates and then create or import from the file system any secrets for the annotations.

It would be pretty simple to do the same thing inside helm; that just before the chart is uploaded we iterate through all the resources in a chart (Replication Controllers really to look at the PodTemplates) and look for these annotations and if so lazily create any secrets. 

Before I submit a PR I just wanted to check if you were happy with this general approach. 

We could leave the secret generation as a separate CLI step in helm; then we could add a 'pre install command' into the charts to install gofabric8 if its not already and then just run `gofabric8 secrets` on the charts before they are uploaded.

Though secret generation sounds like its going to be so common for any vaguely secure chart - and the code is so simple; it does feel like we should just add native support for secret generation/import at chart installation/update time?

technosophos:
If you take a look at `plugins/sec`, we were just starting down the road of making it easy to generate secrets (as a separate step). 

I _really_ like the idea of using annotations to indicate that a secret needs to be generated. It means we don't even have to parse the full RC/Pod to figure out what we need to generate.

We've gone back and forth about hooks like pre/post install, and for the time being we decided to hold back. I would really like to see a separate issue on that so that we can all discuss. We're learning a lot about how Helm uses are trying to use Helm. I think that'd be a great place to get some feedback.

In summary, my view is: 

* +1 on the annotations approach.
* +1 on a command for adding secrets. We should figure out how best to do this.
* And we should discuss pre-install hooks on an issue

Would also like to hear from @gabrtv

[...]

Issue #292, opened 2015-11-26
should we use `install` to upgrade to a new Chart version, or should we have an 'upgrade' command?


jstrachan:
I just realised there wasn't an `upgrade` command (like `brew upgrade`) and wondered if this was intentional.

Is the idea that you re-run `helm install` whenever charts update? e.g.

```
helm install cheese
# time passes and new verisons of charts arrive

helm update
helm install cheese
```

Or should the second `install` fail as its already installed and the `upgrade` command should be used to apply upgrades where you need to specify the policy (e.g. destroy then recreate, merge, rolling upgrade etc). 

```
helm install cheese
# time passes and new verisons of charts arrive

helm update
helm upgrade cheese rolling
```

Relates to #1

jstrachan:
kubectl has options `create`, `update`, `replace`; with a force mode for `update` then for RCs its got `rolling-update` so I wonder if a new `upgrade` command needs to wrap up these different ways of doing an upgrade?

jstrachan:
I've tried implementing this in the openshift-all branch here: https://github.com/fabric8io/helm/tree/openshift-all which has all the PRs merged together. It seems to be working pretty reasonably. 

So you can do
```
helm update
helm upgrade foo
```
which then effectively does a `kubectl apply -f ....` on each resource to update it.

I can cherry pick this to a clean master branch and submit a PR if you fancy it?

jstrachan:
incidentally it seems that `kubectl update` is deprecated and `kubectl replace` is the long term replacement; which only works if the kube resource is complete with its full runtime spec and resourceVersion etc. So its a little tricky to use from helm directly from a chart's resources; we'd have to get the current JSON/YAML and merge in place the resource from the chart. 

So I've switched to using `kubectl apply` which is much more forgiving; though you do need to set the `resourceVersion`. Its working at least ;)

I wonder if its worth renaming `helm upgrade` to `helm apply`? As right now the logic either creates or applies changes depending on if each resource is installed or not. e.g. a new version of a chart might add a new update an existing RC but add a new service or RC; so being able to create or apply on a per resource basis based on whats already installed will be really handy. 

Its just a question of what we call the helm command though ;). 'upgrade' does seem brew-ish at least

technosophos:
This was on our roadmap to figure out, so thank you for jumping in on this. I like the way you've approached it. It feels right from a user perspective.

[...]

PR #293, opened 2015-11-27
first spike of #282 to support annotation based secret generation on …

jstrachan:
…install

### DIFF
--- a/action/generate_secrets.go
+++ b/action/generate_secrets.go
@@ -1,0 +1,234 @@
+package action
+
+import (
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/helm/helm/log"
+
+	"golang.org/x/crypto/ssh"
+
+	"k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/api/unversioned"
+	"k8s.io/kubernetes/pkg/api/v1"
+)
+
+// secretSettings contains the flags to determine how the Secret generation works
+type secretSettings struct {
+	PrintImportFolders  bool
+	WriteGeneratedKeys  bool
+	GenerateSecretsData bool
+}
+
+// keypair is an SSH private and public pair
+type keypair struct {
+	pub  []byte
+	priv []byte
+}
+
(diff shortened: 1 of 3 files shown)


technosophos:
Could we also add a top-level command for this? I can see a use case where people may want to generate the secret, but not run `helm install` (maybe use `kubectl` directly instead). Maybe something like `helm gen-secret` or similar? (I'll leave it up to you.)

I had done some similar work with secrets generation in the `plugins/sec/` folder. The idea was to make a functional example plugin, but one of the capabilities that had was generating random "password" or token strings. Would that be a useful feature for this command? (We can add it later... just thinking out loud.)

Also, would you mind amending the commit message so it follows the repo style? Not a huge deal, but we try to stay consistent for our tooling. https://github.com/deis/deis/blob/master/CONTRIBUTING.md#commit-style-guideline

technosophos:
Sorry... I finally got a big PR merged and it's breaking your stuff.

We changed the way we did things internally so that we were not so reliant on hard-coded Kubernetes kinds. I think this will actually make your work a lot easier. There are fewer cases where you'll need Kind-specific code. I also added some utility logic for reading and modifying labels and annotations. It's not great, but its a start.

technosophos:
We're talking about pushing this feature into the 0.3 milestone, which is scheduled to go out a few weeks from now. Does that give you enough time to massage it into orde,r @jstrachan ?

jstrachan:
@technosophos BTW I hit a few issues with the helm codec stuff (was getting SecurityContext generating an upper case RunAsRoute element so I've been mostly using the kubernetes codec when needing to modify things like an RC. We need to modify the RC when doing `helm upgrade` for #292. We might wanna add some helper methods around working with k8s runtime.Object and kubernetes codec?

e.g. here's the use of the kubernetes codec along with 2 little helper functions to get the ObjectMeta or TypeMeta for the objects so you can modify labels/annotations/type/APIVersion etc
https://github.com/fabric8io/helm/blob/61c8ca7b0ea2a45ff82bf4dd1823196adf54640d/action/install.go#L331-L341

as you need to modify the APIVersion to be able to do a `kubectl apply` from a chart

jstrachan:
I just submitted a fresh PR for #282 after rebasing and cherry picking the secrets changes from PR https://github.com/helm/helm/pull/323 as they were fairly simple. I'll close this PR now

Issue #265, opened 2015-11-20
Make CONTRIBUTING.md stand on its own


arschles:
Right now it says that helm is part of Deis. Thinking it would be better to specify the contributing rules for Helm w/o linking back to Deis (even though the contribution rules are the same)

bacongobbler:
The logic behind this was so we didn't have to maintain multiple identical files for contributing to any deis-related project. We would have one source of truth for contributions. I can see the idea behind this though.

arschles:
@bacongobbler yea, I see the idea behind keeping the contribution rules all in one place as well. I'm ok either way for now, just wanted to keep this on the record

technosophos:
Yeah, we should probably rewrite that file to be a little more generic, even if it's something like "we follow the guidelines set by..." and then point back to Deis

PR #294, opened 2015-11-30
chore(CONTRIBUTING): explain contribution process

technosophos:
Issue #265 suggests that we should have contribution guidelines that are
more specific to the Helm project. This adds a few high-level notes and
then points the user back to the Deis guidelines.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,39 @@
 # How to Contribute
 
-This project is part of Deis. You can find the latest contribution
-guidelines [at the Deis project](https://github.com/deis/deis/blob/master/CONTRIBUTING.md).
+We love getting Pull Requests (PRs). To make sure we keep code quality
+and consistency high, we do have some process in place. In a nutshell:
+
+- Code should follow Go coding standards and pass `go lint` and `go
+  vet`. (These tools are automatically run on every pull request.)
+- PRs should contain a single feature or fix, and follow the conventions
+  linked below.
+- Contributors must agree to the DCO
+- Every patch must be signed off by two core contributors (and this is
+  made easier when the community at large weighs in on PRs, too).
+
+Helm follows the contribution guidelines established by the Deis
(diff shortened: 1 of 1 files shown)


technosophos:
Would love to get a review from @bacongobbler and @mboersma -- and some suggestions as to whether any of this should make its way to the prototype repo.

bacongobbler:
I'd say everything but the `go lint/fmt/vet` comment could be moved upstream. A lot of it is process documentation which is nice but some projects do not require Go, for example deis/postgres :)

CONTRIBUTING.md
```diff
@@ -1,5 +1,38 @@
 # How to Contribute
 
-This project is part of Deis. You can find the latest contribution
-guidelines [at the Deis project](https://github.com/deis/deis/blob/master/CONTRIBUTING.md).
+We love getting Pull Requests (PRs). To make sure we keep code quality
+and consistency high, we do have some process in place. In a nutshell:
+
+- Code should follow Go coding standards and pass `go lint` and `go vet`
```
sgoings: We should also say something about these tools being run automatically on pull requests (which we should set up for those projects that don't have PR builds enabled with anything valuable).

CONTRIBUTING.md
```diff
@@ -1,5 +1,38 @@
...
+- Awaiting review: The PR is ready to be considered for merging. If a PR
+  does not have this, we assume that it's a work in progress (even if
+  the issue does not say WIP)
+- Proposal: something we're discussing, but haven't decided on
+- Enhancement: a feature or chore
+- Bug: a bug
+- Needs manual testing: one of the core team must manually test before
+  LGTM
+- LGTM1: First LGTM (Looks Good To Me)
+- LGTM2: Second LGTM; this means PR can be merged
```
sgoings: Don't we have to merge the PR for them though? maybe: "this means the PR will soon be merged by a project maintainer"?

PR #306, opened 2015-12-03
feat(manifest.go): adding support for yml extensions in manifests

arschles:
replaces #266, as that was having build issues that I have yet to determine the root cause of


arschles:
Fixes #248

PR #322, opened 2015-12-08
ref(*): refactor kubectl to a package

adamreese:
Starting to break up https://github.com/helm/helm/pull/264 into smaller chunks

### DIFF
--- a/kubectl/command.go
+++ b/kubectl/command.go
@@ -1,0 +1,32 @@
+package kubectl
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"os/exec"
+	"strings"
+)
+
+type cmd struct {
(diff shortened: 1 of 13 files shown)


kubectl/command.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"os/exec"
+	"strings"
+)
+
+type Cmd struct {
```
arschles: can you add a docstring for this `struct` and the below `func`s

adamreese: These do not need to be exported.  Pushing an update

arschles: k, giving an LGTM since that was my only comment

Issue #290, opened 2015-11-26
[OpenShift extension] support for OAuthClient


jstrachan:
OpenShift has OAuth support which isn't yet upstreamed to Kubernetes but [hopefully will soon](https://github.com/kubernetes/kubernetes/issues/10408).

Charts for web consoles wishing to reuse OpenShift's OAuth for single sign on with whatever OAuth provider is being used (e.g. the default implementation or github / google et al), need to register an `OAuthClient` resource so that after forwarding to the OAuth provider, the provider can redirect back to the console to implement OAuth properly.

This is a fairly minor change to support the OAuthClient extension resource from OpenShift; which hopefully will move upstream to Kubernetes soon.

jstrachan:
its included in https://github.com/helm/helm/pull/323 now btw

PR #323, opened 2015-12-08
fixes #280, #282, #290, #292 in a single PR

### LINKED ISSUE #280: support OpenShift Templates in Charts
Here's details on OpenShift Templates: https://docs.openshift.org/latest/dev_guide/templates.html

Its basically a simple way to parameterise the JSON/YAML of your kubernetes resources with parameters which have names, default values and expressions to aid in their generation/configuration with environment specific values. 

This relates to #108.

Note that OpenShift Templates are [hopefully going to be pushed upstream](https://github.com/kubernetes/kubernetes/issues/10408) into Kubernetes; they are fairly small simple things; basically [an array of Parameters and an array of JSON/YAML templates](https://github.com/openshift/origin/blob/master/pkg/template/api/v1/types.go#L10-L24) using simple `${PARAMETER_NAME}` expressions to replace values into the kubernetes resources.

```
type Template struct {
    unversioned.TypeMeta `json:",inline"`
    kapi.ObjectMeta      `json:"metadata,omitempty"`

    // Objects is an array of objects to include in this template. Required.
    Objects []runtime.RawExtension `json:"objects" description:"list of objects to include in the template"`

    // Optional: Parameters is an array of Parameters used during the
    // Template to Config transformation.
    Parameters []Parameter `json:"parameters,omitempty" description:"optional: list of parameters used during template to config transformation"`

    // Labels is a set of labels that are applied to every
    // object during the Template to Config transformation. Optional
    Labels map[string]string `json:"labels,omitempty" description:"optional: list of lables that are applied to every object during the template to config transformation"`
}
```

A PR is coming shortly along with a chart repository with lots of examples.

I think when [Config Data](https://github.com/kubernetes/kubernetes/blob/master/docs/proposals/config_data.md) arrives upstream the need for OpenShift Templates is less great; OpenShift templates are a similar idea just without the `Config Data` being stored and managed inside kubernetes; instead the effective config data is replaced into the OpenShift templates before the resources are created. 

From a Helm perspective Config Data is going to be easier for helm to support nicely as kubernetes will manage the Config Data and inject it into volumes or environment variables; so there's less for helm to do ;)

This effort is mostly an experiment to use helm to install and update the charts created in the [fabric8 project](http://fabric8.io/) for things like [CI / CD](http://fabric8.io/guide/cdelivery.html), [ChatOps](http://fabric8.io/guide/chat.html) and [Chaos Monkey](http://fabric8.io/guide/chaosMonkey.html) on top of kubernetes.

jstrachan:
lots of these fixes involve adding parameters to the same lines so its kinda hard to keep them app rebased and separate so here's a PR of all the changes in one go. 

If this is too much on one PR I'll maybe try cherry pick out a separate issue at a time and when thats merged, cherry pick another etc as its lots of work keeping all these separate branches rebased

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ bin/
 _dist/
 _scripts/ci/bintray-ci.json
 
+.idea
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ endif
 
 BIN_DIR := bin
 DIST_DIR := _dist
-GO_PACKAGES := action chart config dependency log manifest release plugins/sec plugins/example codec
+GO_PACKAGES := action chart config dependency log manifest parameters release plugins/sec plugins/example codec
 MAIN_GO := helm.go
 HELM_BIN := $(BIN_DIR)/helm
 
--- a/action/generate_secrets.go
+++ b/action/generate_secrets.go
@@ -14,15 +14,17 @@ import (
 
 	"golang.org/x/crypto/ssh"
 
-	"k8s.io/kubernetes/pkg/api"
 	"k8s.io/kubernetes/pkg/api/unversioned"
 	"k8s.io/kubernetes/pkg/api/v1"
 )
 
+
+const defaultAPIVersion = "v1"
+
 // secretSettings contains the flags to determine how the Secret generation works
 type secretSettings struct {
-	PrintImportFolders bool
-	WriteGeneratedKeys bool
+	PrintImportFolders  bool
+	WriteGeneratedKeys  bool
 	GenerateSecretsData bool
 }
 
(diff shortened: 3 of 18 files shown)


mboersma:
This PR is five months old and needs a rebase. New development is now encouraged at https://github.com/kubernetes/helm instead.

This is great stuff though @jstrachan. Should it be refreshed and re-reviewed for "helm classic" or is it relevant to next-gen [`helm`](https://github.com/kubernetes/helm)?

mboersma:
Helm development continues at https://github.com/kubernetes/helm. This is Helm Classic, which is in maintenance mode.

This PR hasn't been updated in five months, so I'm closing it. Please re-open it if these changes are backward-compatible and can be rebased. Thanks again, and please consider making similar improvements to the [new helm](https://github.com/kubernetes/helm)!

PR #331, opened 2015-12-08
ref(install): refactor install to use kubectl pkg

adamreese:


action/install_test.go
```diff
@@ -0,0 +1,93 @@
...
+	tmpHome := createTmpHome()
+	defer os.RemoveAll(tmpHome)
+	fakeUpdate(tmpHome)
+
+	for _, tt := range tests {
+		defer func() {
+			if err := recover(); err != nil {
+				output.WriteString(err.(string))
+			}
+		}()
```
adamreese: Not sure if I setup this defer correctly

technosophos: Your intuition is right. Defers shouldn't go in for loops.

adamreese: bumped it out of loop

PR #335, opened 2015-12-09
Scanner buffer overwrite bug

givia:
`scanner.Bytes()` does no allocation. By appending its result to a new empty byte array, we're creating a copy to avoid the overwrite problem.


technosophos:
Excellent catch! Thanks -- and thanks for improving the tests, too. As soon as we get an LGTM2, I'll merge.

We follow a consistent commit message format. Not a huge deal, but if you have a moment to amend the commits that would be awesome.

https://github.com/helm/helm/blob/master/CONTRIBUTING.md

bacongobbler:
also, was @remohammadi working with you on this? We should probably ask for his permission to use his commit if he's not explicitly working with you. :)

remohammadi:
I'm working with @givia :) The test data comes from the [etcd project](https://github.com/coreos/etcd/blob/master/hack/kubernetes-deploy/etcd.yml).

bacongobbler:
perfect. thanks!

Issue #309, opened 2015-12-03
PROPOSAL: `helm generate` and built-in generator/template/expansion support


technosophos:
**tl;dr** This proposal suggests adding a `helm generate` tool analogous to the `go generate` tool, with the express purpose of providing generic support for tools that modify a manifest.

There has been a lot of talk about how we can make it easier to do certain generative steps during a various Helm phases. The ecosystem for this is growing, too. I've been trying to figure out a way to extend the plugin system to allow for this. And I'm leaning toward a method based on `go generate`.

## Use cases

As a user, I would like to...

- Generate secrets automatically at fetch or install time
- Modify environment variables (Pods, DSes, RCs, etc) at fetch or install time
- Do full templating client-side
- Do full templating server-side
- Perform simple substitutions
- Generate a manifest from some external source
- Run zero or more different tools against a manifest between fetch and install time
- Track changes to manifests locally (and VCS them)

As a Chart author, I would like to...

- Require that a user have an external tool/plugin before fetch/install
- _Allow_ a user to use a tool/plugin, but fall back on default manifests otherwise
- Create my own tool and use it in charts

## Extending the Plugin System to Allow Generating

The plugin system that exists in Helm today follows the model of Git: If `helm` receives a command that it does not know (`helm foo`), it will search on `$PATH` for a tool matching a particular nameing convention (`helm-foo`). If it finds one, it will set a number of environment variables, and then execute that command, passing it all of the remaining command line options.

This proposal suggests that we can follow a pattern like Go's `go generate` tool and use some of the features of the plugin system.

### The `helm generate` command

The command `helm generate CHART_NAME` will run a series of operations on the given chart in the workspace:

- Open each file in the `manifests/` directory
- Scan for any number of headers of the form `#helm:generate COMMAND [OPTIONS]`
- For each such located header, it will...
   - set up a number of environment variables (like Helm plugins)
   - perform shell expansion on OPTIONS, making OPTIONS'
   - and then execute the supplied COMMAND OPTIONS' inside of a shell.
   - trap and report any errors

## Modifications to `helm fetch` and `helm install`

The `helm fetch` and `helm install` commands will gain a new flag: `--generate, -g`, which will send the chart through the generation system as follows:

- `helm fetch` will copy the chart from the cache to the workspace, and then run the generator
- `helm install` will copy the chart from the cache _if necessary_, run the generator, and (if no errors occurred during generation) run `kubectl create`.

## Open Questions

[...]

PR #339, opened 2015-12-10
feat(generator): add generator tool

technosophos:
This demonstrates the generator functionality as described in
issue #309. The helm-tpl plugin uses Go's built-in template system
along with the Sprig library of helper functions.


generator/generator.go
```diff
@@ -0,0 +1,163 @@
...
+	"strings"
+
+	"github.com/helm/helm/log"
+)
+
+const GeneratorKeyword = "helm:generate "
+
+// Walk walks a chart directory and executes generators as it finds them.
+//
+// Returns the number of generators executed.
```
arschles: can you add some docs on when errors will be returned, and what some of the common ones are?

generator/generator.go
```diff
@@ -0,0 +1,163 @@
...
+	if len(args) == 0 {
+		return errors.New("empty command")
+	}
+	name := args[0]
+	args = args[1:]
+
+	cmd := exec.Command(name, args...)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	cmd.Stdin = os.Stdin
```
arschles: wouldn't it make sense to set the above env variables in `cmd.Env` instead of in the above `os.Setenv`s?

technosophos: I'm actually not sure. The environment variables are used by:

- the generate command itself, which runs substitutions using ExpandEnv
- the generator plugin, which uses environment variables to find out about the environment
- and possibly children of the plugin, which may use those variables for, say, substitution

Since `os.ExpandEnv`, which runs in-process, doesn't seem to be setable to the environment of the `cmd.Env`, I don't think switching would work.

technosophos:
Left to do on this particular ticket:

- [ ] Do a few real-world templates to validate assumptions
- [ ] Add a simple token generator and Base64 support (probably via Sprig)
- [ ] Document the new features

Problems outside of the scope of this PR:

- Generating the initial values files. See the Deis charts repo for examples.
- Adding `fetch --generate` and `install --generate`, which can be done later.
- Adding support for prompting the user to enter a value, which can be done as a separate PR.

technosophos:
@sgoings I am going to need help on this one. `helm` calls itself, expecting that `helm` is on `$PATH`. But it appears that it's not.

sgoings:
This should fix you up:

```
diff --git i/Makefile w/Makefile
index 81a9f9e..ab20e9d 100644
--- i/Makefile
+++ w/Makefile
@@ -7,6 +7,7 @@ DIST_DIR := _dist
 GO_PACKAGES := action chart config dependency log manifest release plugins/sec plugins/example codec
 MAIN_GO := helm.go
 HELM_BIN := $(BIN_DIR)/helm
+PATH_WITH_HELM = PATH="$(shell pwd)/$(BIN_DIR):$(PATH)"

 VERSION := $(shell git describe --tags --abbrev=0 2>/dev/null)+$(shell git rev-parse --short HEAD)

@@ -53,10 +54,10 @@ else
 endif

 quicktest:
-       go test -short ./ $(addprefix ./,$(GO_PACKAGES))
+       $(PATH_WITH_HELM) go test -short ./ $(addprefix ./,$(GO_PACKAGES))

 test: test-style
-       go test -v ./ $(addprefix ./,$(GO_PACKAGES))
+       $(PATH_WITH_HELM) go test -v ./ $(addprefix ./,$(GO_PACKAGES))

 test-style:
        @if [ $(shell gofmt -e -l -s *.go $(GO_PACKAGES)) ]; then \
```

Makefile
```diff
@@ -7,6 +7,7 @@ DIST_DIR := _dist
 GO_PACKAGES := action chart config dependency log manifest release plugins/sec plugins/example codec
 MAIN_GO := helm.go
 HELM_BIN := $(BIN_DIR)/helm
+PATH_WITH_HELM = PATH="$(shell pwd)/$(BIN_DIR):$(PATH)"
```
arschles: Can this just use `${PWD}`?

adamreese:
Test depends on the external binary `helm` in path.  Could blowup when running tests locally if `installed helm` is different than `helm under test`.

action/template_test.go
```diff
@@ -0,0 +1,49 @@
...
+	if out.String() != "Hello World!\n" {
+		t.Errorf("Expected Hello World!, got %q", out.String())
+	}
+
+	// No data
+	out.Reset()
+	Template("", tpl, "")
+	if out.String() != "Hello Clowns!\n" {
+		t.Errorf("Expected Hello Clowns!, got %q", out.String())
+	}
```
arschles: Looks like this test would probably be better written as [table-driven](https://github.com/golang/go/wiki/TableDrivenTests). Looks like columns could be test type (for debugging), filename, and expected rendered value of template.

Not blocking for this PR, something to keep in mind.

adamreese:
Sorry, just saw your note on the external binary issue...

cli/template.go
```diff
@@ -0,0 +1,59 @@
...
+- TOML: .toml
+- JSON: .json
+
+If an output file is specified, the results will be written to the output
+file instead of STDOUT. Writing to the source template file is unsupported.
+(In other words, don't set the source and output to the same file.)
+`
+
+// tplCmd is the command to handle templating.
+// glide tpl -o dest.txt -d data.toml my_template.tpl
```
arschles: `glide tpl`?

docs/generate-and-template.md
```diff
@@ -0,0 +1,294 @@
...
+
+Note that the above variable is the absolute path to the chart
+directory, not to the file that `helm generate` found.
+
+The following guidelines should be kept in mind when writing Helm
+generator plugins:
+
+- All of the named environment variables are available to the plugin.
+- The plugin does _not_ execute in a shell. Thus you cannot expect that
+  implementors will be able to easily redirect STDOUT or STDERR
```
arschles: Does helm redirect plugin stdout/stderr somewhere? Should it?

generator/generator.go
arschles: thanks :+1:

generator/generator.go
arschles: Got it, didn't realize the first bullet point

generator/generator.go
technosophos: This was fixed.

Makefile
technosophos: @sgoings would have to answer that. It's here for TravisCI... so I don't know for sure.

cli/template.go
technosophos: ROTFL... too many package managers.

docs/generate-and-template.md
technosophos: It redirects the stdout/stderr to the current STDOUT/STDERR that the app executes in. So, for example, `helm generate | grep "bar"` will capture any case where a plugin emitted `bar`.

The main issue I was trying to convey here is that you can't do:

```
helm:generate foo > bar
```

And that's really because `foo > bar` does not execute in a shell.

Makefile
arschles: k

generator/generator.go
arschles: Thinking about this a little more, it might be wiser to use [os.Expand](https://godoc.org/os#Expand) instead of `os.ExpandEnv`. Doing so would accomplish all three of the above bullet points and not require any changes to the process environment, only changes to the command's (which its children can use as well).

It's a cleaner solution IMO, but I wouldn't say it's required for this PR.

arschles:
@technosophos :+1:

adamreese:
Is there an issue created for the isolating the tests from user `PATH`?

technosophos:
@adamreese what do you mean? I'm not sure what the contents of that ticket would be.

Makefile
sgoings: Yeah it probably can, @arschles.

arschles: k. I guess either is fine, I suggested `${PWD}` because afaik it doesn't invoke the shell

Issue #286, opened 2015-11-25
helm edit is not saving


technosophos:
I tried the following against HEAD:

```
⇒  export EDITOR=mvim                                                                                                                                                                                   
☃ @ENG001590.local ~/Code/Go/src/github.com/helm/helm «/Users/mattbutcher/Code» ‹feat/objects-1.1*›
⇒  helm edit openldap
```

Inside of the editor, I modified the Chart.yaml to have lots of details, setting description, home, maintainers and details. Then I did `:wq` in VIM.

When I went to my `$HELM_HOME/workspace/charts/openldap` chart, the Chart.yaml was set back to its originally generated value:

```yaml
name: openldap
home: http://example.com/your/project/home
version: 0.1.0
description: Provide a brief description of your application here.
maintainers:
- Your Name <email@address>
details: |-
  This section allows you to provide additional details about your application.
  Provide any information that would be useful to users at a glance.
```

/cc @gabrtv

technosophos:
Over a couple of chats, we think we've isolated this to the case where the editor is forked off of the shell (e.g. gvim or mvim instead of regular vim).

adamreese:
We could have the editor open the chart directory directly like homebrew.  Then saves would be directly on the files

PR #348, opened 2015-12-14
bug(edit): make editor open chart folder directly

adamreese:
Fixes: https://github.com/helm/helm/issues/286

This PR changes the editing workflow by opening the chart directory in `$EDITOR` rather than generating a single temp file for all manifests.

### DIFF
--- a/action/edit.go
+++ b/action/edit.go
@@ -1,22 +1,12 @@
 package action
 
 import (
-	"bytes"
-	"fmt"
-	"io/ioutil"
 	"os"
 	"os/exec"
 	"path"
-	"path/filepath"
-	"regexp"
-)
 
-import (
 	"github.com/helm/helm/log"
 )
-
-// delimeterRegexp captures relative path and file content
-var delimeterRegexp = regexp.MustCompile(`--- # (.+)\n([\s\S]*?)--- # end\n`)
 
 // Edit charts using the shell-defined $EDITOR
 //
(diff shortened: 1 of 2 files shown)


mboersma:
When I quit `vi` on Mac OS X, I see an error each time:
```console
$ set | grep EDITOR
EDITOR=vi
$ helm edit deis  # blocks until editing ends, after which you see...
[ERROR] Could not open $EDITOR: exit status 1
$ export EDITOR=atom
$ helm edit deis  # command returns immediately
```

adamreese:
weird, looks like it is only `vi`.  `vim` seems to work fine

technosophos:
That's a known issue with `vi`. The fix is to specify a full path on $EDITOR:
http://stackoverflow.com/questions/4092432/could-not-execute-editor

adamreese:
didn't seem to fix it

mboersma:
Yes, it works fine with `EDITOR=vim`, but the StackOverflow workaround of `EDITOR=/usr/bin/vi` still gives an error. `vim` is fine with me as a workaround, and if this is the only editor config where this happens then it's probably fine to :ship:.

technosophos:
I suspect that it has something to do with the fact that `vi` is a symlink.

```
⇒  ls -lah /usr/bin/vi
lrwxr-xr-x  1 root  wheel     3B Apr 11  2015 /usr/bin/vi -> vim
```

adamreese:
@mboersma does that count as a LGTM?

technosophos:
Oh, wait... I can't LGTM twice.

PR #352, opened 2015-12-15
ref(tests): use test helpers in all unit tests

adamreese:
/cc @lanej


lanej:
:+1:

technosophos:
@lanej does :+1: constitute an LGTM?

PR #355, opened 2015-12-16
feat(lint): lint by path

lanej:
* accepts reference to chart in home directory or path
* remove useless flag test
* Fixes #311

PR #359, opened 2015-12-17
encapsulate path formation

lanej:


lanej:
@technosophos @adamreese this is what i had in mind per our coversation in #355

util/structure.go
```diff
@@ -0,0 +1,21 @@
+package util
+
+import "path/filepath"
+
+// CachePath is the suffix for the cache.
+const CachePath = "cache"
+
+// CacheChartPath is the directory that contains a user's cached charts.
+const CacheChartPath = "cache/charts"
```
adamreese: `CacheChartPath` can probably get axed.  We now use `cache/$repo`

adamreese:
I like it!

technosophos:
Yeah.

technosophos:
Wow. The diff pretty much proves that you hit a valuable abstraction point.

PR #361, opened 2015-12-18
fix(lint): WARN on failed checks

lanej:


sgoings:
We probably want to figure out which ones are errors and which ones are warnings... but I like this approach for now.

lanej:
@sgoings i can turn the warnings slice into a map of level and messages.  will do that in a different PR. this will at least allow the charts repo to pass CI.

PR #371, opened 2016-01-04
ref(*): move validation out of action package

michelleN:


arschles:
@michelleN and I talked offline and it looks like the now-tiny `validation` package will be growing. I believe that splitting it out now is a good design decision.

Issue #325, opened 2015-12-08
repo add fails on a new helm setup without running update first


helgi:
Blew away my helm setup and got myself a new binary et all but didn’t run `update` by accident

 ```
helm repo add deis https://github.com/deis/charts
[ERROR] Could not load /Users/helgi/.helm/config.yaml: open /Users/helgi/.helm/config.yaml: no such file or directory
```

This could show a bit nicer error message or just have certain actions run the eq of `update` if needed.

michelleN:
Yeah, we can definitely run helm doctor to fix the set up before failing this way.

mboersma:
This was fixed by #377. `helm repo add` does the right thing now.

PR #377, opened 2016-01-07
ref(repo): run helm doctor before failing

michelleN:
This resolves issue #325.


jackfrancis:
FYI, a helper function might be a good idea if this ever gets more complicated (see the duplicate "load config and do something that errors out" statements inside `mustConfig`.

But currently that would add unnecessary complexity, so LGTM.

michelleN:
thanks ya'll!

PR #384, opened 2016-01-13
Use bash to install helm

2opremio:
Doing `curl -s https://get.helm.sh | sh` results in `sh: 13: set: Illegal option -o pipefail`. I guess the installation script is not POSIX-compliant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ $ helm repo add charts https://github.com/helm/charts
 
 From a Linux or Mac OS X client:
 ```
-curl -s https://get.helm.sh | sh
+curl -s https://get.helm.sh | bash
 ```
 
 *or*:


sgoings:
Why can't we just fix the install script to be POSIX compliant?

krancour:
> Why can't we just fix the install script to be POSIX compliant?

+1

2opremio:
> Why can't we just fix the install script to be POSIX compliant?

That would be even better. 

But note that http://helm.sh already reads `curl -s https://get.helm.sh | bash` so something tells me that using bash was intentional.

sgoings:
We have the ability to determine our own destiny, @2opremio! :wink:

sgoings:
The installer is located here: https://github.com/helm/helm/blob/master/_scripts/installer/install.sh if you're interested in taking a crack at it. I believe the deployment of that script is manual right now but I can fix that.

rimusz:
We had already a discussion about http://helm.sh using `curl -s https://get.helm.sh | bash` instead of `sh` and `bash` won. 
We can merge this PR to have consistency between helm.sh site and github repo.
And if somebody wants to make the install script to be POSIX compliant sure why not.

bacongobbler:
LGTM for now as a doc fix. If someone wants to take a crack at making the script POSIX-compliant then that would be awesome, but for now let's stick with what works.

Also note that we use this same script to install the Deis client as well, so any changes made to the helm script could be ported over to the Deis client install script.

PR #390, opened 2016-02-01
Decoupled templating functionality from file IO

colonelmo:
Templates could've only been generated using data from files and the output had to be written to a file to.
GenerateTemplate now allows one to generate templates from and to io.Reader and io.Writer objects respectively.
This can come in handy in situations where the data is generated/manipulated dynamically (say, through a web ui).

### DIFF
--- a/action/template.go
+++ b/action/template.go
@@ -15,6 +15,18 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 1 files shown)


technosophos:
@adamreese or @michelleN  Can we get another code review on this? I'd like to merge it for the next release.

@colonelmo Thanks for this. And I am sorry for the delay in responding. That was my fault.

PR #395, opened 2016-02-18
ref(glide): Using glide 0.9.0 new import resolver

mattfarina:
- `glide up` run with 0.9.0 that walks the import tree rather than
  the filesystem tree. Pulls just the 36 dependencies in the tree.
- Removes travic ci caching. The number of packages being
  is small enough fetching them should be faster than copying from
  the travis cache.
- The glide install run Make bootstrap should be backwards
  compatible to glide 0.8.3.

The output of `glide up` is at https://gist.github.com/mattfarina/009fad1cfbaccdce7aca

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,17 +8,13 @@ branches:
     - master
     - /^v?(?:[0-9]+\.){2}[0-9]+.*$/
 
-cache:
-  directories:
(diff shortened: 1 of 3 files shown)


technosophos:
I'm doing some manual testing, and then will merge.

PR #418, opened 2016-03-22
feat(containerize helm build): Makefile supporting containerized build

chaitanyaenr:


Makefile
```diff
@@ -1,28 +1,79 @@
...
 
+REPO_PATH := github.com/helm/${SHORT_NAME}
+
+# The following variables describe the containerized development environment
+# and other build options
+DEV_ENV_IMAGE := quay.io/deis/go-dev:0.9.0
+DEV_ENV_WORK_DIR := /go/src/${REPO_PATH}
+DEV_ENV_CMD := docker run --rm -v ${CURDIR}:${DEV_ENV_WORK_DIR} -w ${DEV_ENV_WORK_DIR} ${DEV_ENV_IMAGE}
+DEV_ENV_CMD_INT := docker run -it --rm -v ${CURDIR}:${DEV_ENV_WORK_DIR} -w ${DEV_ENV_WORK_DIR} ${DEV_ENV_IMAGE}
+LDFLAGS := "-s -X main.version=${VERSION}"
```
arschles: Not sure, but does this need to go under the line that declares `VERSION`?

Makefile
```diff
@@ -1,28 +1,79 @@
...
+# The following variables describe the source we build from
+GO_FILES := $(wildcard *.go)
+GO_DIRS := model/ nginx/ utils/ utils/modeler
+GO_PACKAGES := ${REPO_PATH} $(addprefix ${REPO_PATH}/,${GO_DIRS})
 
-export GO15VENDOREXPERIMENT=1
+# The following variables describe k8s manifests we may wish to deploy
+# to a running k8s cluster in the course of development.
+RC := manifests/deis-${SHORT_NAME}-rc.yaml
+SVC := manifests/deis-${SHORT_NAME}-service.yaml
```
arschles: these lines are unneeded

Makefile
```diff
@@ -1,28 +1,79 @@
...
 DIST_DIR := _dist
 GO_PACKAGES := action chart config dependency log manifest release plugins/sec plugins/example codec
 MAIN_GO := helm.go
 HELM_BIN := $(BIN_DIR)/helm
 PATH_WITH_HELM = PATH="$(shell pwd)/$(BIN_DIR):$(PATH)"
-
 VERSION := $(shell git describe --tags --abbrev=0 2>/dev/null)+$(shell git rev-parse --short HEAD)
+# The following variables describe the source we build from
+GO_FILES := $(wildcard *.go)
+GO_DIRS := model/ nginx/ utils/ utils/modeler
```
arschles: This line is unneeded

Makefile
```diff
@@ -1,28 +1,79 @@
...
-endif
+# Allow developers to step into the containerized development environment
+dev: check-docker
+	${DEV_ENV_CMD_INT} bash
+
+dev-registry: check-docker
+	@docker inspect registry >/dev/null 2>&1 && docker start registry || docker run --restart="always" -d -p 5000:5000 --name registry registry:0.9.1
+	@echo
+	@echo "To use a local registry for Deis development:"
+	@echo "    export DEIS_REGISTRY=`docker-machine ip $$(docker-machine active 2>/dev/null) 2>/dev/null || echo $(HOST_IPADDR) `:5000/"
```
arschles: I don't believe this build target is needed

Makefile
```diff
@@ -31,34 +78,36 @@ build-all:
...
+		kubectl delete rc deis-${SHORT_NAME} --namespace=deis; \
+		kubectl create -f manifests/deis-${SHORT_NAME}-rc.tmp.yaml; \
+	else \
+		kubectl create -f manifests/deis-${SHORT_NAME}-rc.tmp.yaml; \
+	fi
 
-test: test-style
-	$(PATH_WITH_HELM) go test -v ./ $(addprefix ./,$(GO_PACKAGES))
+examples:
+	kubectl create -f manifests/examples.yaml
```
sgoings: Pretty sure this path doesn't exist in this repository...

Makefile
chaitanyaenr: @arschles done

Makefile
chaitanyaenr: @arschles done

Makefile
chaitanyaenr: @arschles done

Makefile
```diff
@@ -31,34 +78,36 @@ build-all:
...
-quicktest:
-	$(PATH_WITH_HELM) go test -short ./ $(addprefix ./,$(GO_PACKAGES))
+deploy: check-kubectl dev-release
+	@kubectl describe rc deis-${SHORT_NAME} --namespace=deis >/dev/null 2>&1; \
+	if [ $$? -eq 0 ]; then \
+		kubectl delete rc deis-${SHORT_NAME} --namespace=deis; \
+		kubectl create -f manifests/deis-${SHORT_NAME}-rc.tmp.yaml; \
+	else \
+		kubectl create -f manifests/deis-${SHORT_NAME}-rc.tmp.yaml; \
+	fi
```
sgoings: This target doesn't belong in helm.

Makefile
```diff
@@ -31,34 +78,36 @@ build-all:
...
-# TRAVIS_TAG is set to the tag name if the build is a tag
-ifdef TRAVIS_TAG
-	@jq '.version.name |= "$(VERSION)"' _scripts/ci/bintray-template.json | \
-		jq '.package.repo |= "helm"' > _scripts/ci/bintray-ci.json
-else
-	@jq '.version.name |= "$(VERSION)"' _scripts/ci/bintray-template.json \
-		> _scripts/ci/bintray-ci.json
-endif
+set-image:
+	sed "s#\(image:\) .*#\1 ${IMAGE}#" manifests/deis-${SHORT_NAME}-rc.yaml > manifests/deis-${SHORT_NAME}-rc.tmp.yaml
```
sgoings: This target doesn't belong in helm.

Makefile
chaitanyaenr: @arschles done

Makefile
chaitanyaenr: @sgoings removed

Makefile
```diff
@@ -1,28 +1,61 @@
...
 build: $(MAIN_GO)
 	go build -o $(HELM_BIN) -ldflags "-X github.com/helm/helm/cli.version=${VERSION}" $<
 
-bootstrap:
-	go get -u github.com/golang/lint/golint github.com/mitchellh/gox
-	glide install
+install: check-docker
+	${DEV_ENV_CMD} make build
+	${DEV_ENV_CMD} install -d ${DESTDIR}/usr/local/bin/
+	${DEV_ENV_CMD} install -m 755 $(HELM_BIN) ${DESTDIR}/usr/local/bin/helm
```
arschles: Are these lines supposed to be `go install`? Also, `/usr/local/bin` won't be mounted to the host in this and the above command.

Makefile
```diff
@@ -1,28 +1,61 @@
...
+docker-build: build check-docker
+	docker build --rm -t ${IMAGE} rootfs
+	docker tag -f ${IMAGE} ${MUTABLE_IMAGE}
+
+test-style: check-docker
+	${DEV_ENV_CMD} make test-style
+
+# Builds the binary-- this should only be executed within the
+# containerized development environment.
+
```
arschles: can you remove this line, so the comment is right above the target

Makefile
chaitanyaenr: @sgoings removed

Makefile
```diff
@@ -31,33 +64,23 @@ build-all:
...
-# TRAVIS_TAG is set to the tag name if the build is a tag
-ifdef TRAVIS_TAG
-	@jq '.version.name |= "$(VERSION)"' _scripts/ci/bintray-template.json | \
-		jq '.package.repo |= "helm"' > _scripts/ci/bintray-ci.json
-else
-	@jq '.version.name |= "$(VERSION)"' _scripts/ci/bintray-template.json \
-		> _scripts/ci/bintray-ci.json
-endif
+set-image:
+	sed "s#\(image:\) .*#\1 ${IMAGE}#" manifests/deis-${SHORT_NAME}-rc.yaml > manifests/deis-${SHORT_NAME}-rc.tmp.yaml
```
arschles: don't need this build tag

Makefile
```diff
@@ -70,6 +93,16 @@ test-style:
 		go vet github.com/helm/helm/$$i; \
```
arschles: this command should be done inside a container

Makefile
chaitanyaenr: @sgoings changes done

Makefile
```diff
@@ -1,28 +1,61 @@
...
+# Allow developers to step into the containerized development environment
+
+# Containerized dependency resolution
+bootstrap: check-docker
+	${DEV_ENV_CMD} glide install
+
+# Containerized build of the binary
+dist: check-docker
+	mkdir -p ${BIN_DIR}
+	${DEV_ENV_CMD} make binary-build
```
arschles: `binary-build` is already run inside a container, so this command will fail. it should run the target, not in a container

chaitanyaenr: @arschles changes done

Makefile
```diff
@@ -1,83 +1,61 @@
...
-endif
-
-quicktest:
-	$(PATH_WITH_HELM) go test -short ./ $(addprefix ./,$(GO_PACKAGES))
+build-all:
+	gox -verbose \
+	-ldflags "-X github.com/helm/helm/cli.version=${VERSION}" \
+	-os="linux darwin " \
+	-arch="amd64 386" \
+	-output="$(DIST_DIR)/{{.OS}}-{{.Arch}}/{{.Dir}}" .
```
arschles: this needs to be done in a container

Makefile
arschles: @chaitanyaenr this should be removed

Makefile
```diff
@@ -1,28 +1,59 @@
...
+# containerized development environment.
+
+binary-build:
+	make build-all
+	@cd $(DIST_DIR) && \
+	find * -type d -exec zip -jr helm-$(VERSION)-{}.zip {} \; && \
+	cd -
 
 build-all:
 	gox -verbose \
```
arschles: @chaitanyaenr don't forget, this still needs to be done inside a container. For example, `${DEV_ENV_CMD} gox ...`

arschles:
Still TODO:

- [ ] Create a new `go-dev` image (in https://github.com/deis/docker-go-dev) that includes `gofmt`

chaitanyaenr:
@arschles issue created and assigned.

.travis.yml
```diff
@@ -15,6 +15,10 @@ cache:
 go:
   - 1.5.1
 
+sudo: required
+services:
+- docker
+
```
krancour: * `language` should be `generic`; no longer go (since all the go tools are in a container).
* Do not need to set `GO15VENDOREXPERIMENT=1`.  This is already set inside the container.
* Do not need to specify go version.
* All install steps can go away.  Those tools all exist already within the container.  You can replace with a _single_ install step: `make bootstrap`
* `script` should be just `make build test`.  This will run a containerized build for Linux/64 within the container and the tests will be executed within the container, against the code _and_ the binary.
* Everything else looks correct.

Makefile
```diff
@@ -1,47 +1,63 @@
-ifndef GOPATH
-$(error No GOPATH set)
-endif
+SHORT_NAME := helm
+DEIS_REGISTRY ?= ${DEV_REGISTRY}
+IMAGE_PREFIX ?= helm
```
krancour: `DEIS_REGISTRY` and `IMAGE_PREFIX` are not needed.  These are only relevant when a build is producing images that must be push to a Docker registry.  That's not applicable here since this build only produces binaries.

Makefile
```diff
@@ -1,47 +1,63 @@
-ifndef GOPATH
-$(error No GOPATH set)
-endif
+SHORT_NAME := helm
```
krancour: This variable is only used in one place.  Get rid of it and paste it's value in the one place it is used.

Makefile
krancour: The correct `LDFLAGS` for this build are:
```
LDFLAGS := "-X ${REPO_PATH}/cli.version=${VERSION}"
```

Makefile
```diff
@@ -1,47 +1,63 @@
...
 PATH_WITH_HELM = PATH="$(shell pwd)/$(BIN_DIR):$(PATH)"
-
 VERSION := $(shell git describe --tags --abbrev=0 2>/dev/null)+$(shell git rev-parse --short HEAD)
+DEV_ENV_IMAGE := quay.io/deis/go-dev:0.9.0
+DEV_ENV_WORK_DIR := /go/src/${REPO_PATH}
+DEV_ENV_CMD := docker run --rm -v ${CURDIR}:${DEV_ENV_WORK_DIR} -w ${DEV_ENV_WORK_DIR} ${DEV_ENV_IMAGE}
+DEV_ENV_CMD_INT := docker run -it --rm -v ${CURDIR}:${DEV_ENV_WORK_DIR} -w ${DEV_ENV_WORK_DIR} ${DEV_ENV_IMAGE}
+LDFLAGS := "-s -X main.version=${VERSION}"
 
-export GO15VENDOREXPERIMENT=1
```
krancour: This shouldn't go away.  It's already set inside the container, but there should be _one_ step like `binary-build` (which I think we should rename to `native-build`) that _might_ be executed outside the container if a developer really wants to build helm from source for their laptop's native OS and architecture _only_.  For the sake of that one edge case, this variable should remain.

Makefile
```diff
@@ -1,47 +1,63 @@
...
 
-ifndef VERSION
-  VERSION := git-$(shell git rev-parse --short HEAD)
-endif
 
-build: $(MAIN_GO)
-	go build -o $(HELM_BIN) -ldflags "-X github.com/helm/helm/cli.version=${VERSION}" $<
+# Allow developers to step into the containerized development environment
```
krancour: There should be a target here.  Something like:

```
dev: check-docker
	${DEV_ENV_CMD_INT} bash
```

Makefile
```diff
@@ -1,47 +1,63 @@
...
+
+check-docker:
+	@if [ -z $$(which docker) ]; then \
+		echo "Missing \`docker\` client which is required for development"; \
+		exit 2; \
+	fi
+# Containerized build of the binary
+build: check-docker
+	mkdir -p ${BIN_DIR}
+	make binary-build
```
krancour: This step should be containerized, but it isn't.

Makefile
```diff
@@ -1,47 +1,63 @@
...
+
+docker-build: build check-docker
+	docker build --rm -t ${IMAGE} rootfs
+	docker tag -f ${IMAGE} ${MUTABLE_IMAGE}
```
krancour: This target isn't applicable.  We're not building Docker images; only binaries.

Makefile
```diff
@@ -1,47 +1,63 @@
...
 
-build-all:
-	gox -verbose \
-	-ldflags "-X github.com/helm/helm/cli.version=${VERSION}" \
-	-os="linux darwin " \
-	-arch="amd64 386" \
-	-output="$(DIST_DIR)/{{.OS}}-{{.Arch}}/{{.Dir}}" .
 
-clean:
-	rm -rf $(DIST_DIR) $(BIN_DIR)
```
krancour: This `clean` target shouldn't be removed.  `clean` has been rewritten farther below, but the implementation is wrong.

Makefile
```diff
@@ -1,47 +1,63 @@
...
+# Builds the binary-- this should only be executed within the
+# containerized development environment.
 
-dist: build-all
+binary-build:
```
krancour: Might be more clear if we rename this to `native-build`, since that's an accurate description of what it does.  It compiles for the native OS and architecture wherever it is run.

Makefile
```diff
@@ -1,47 +1,63 @@
...
+	make build-all
 	@cd $(DIST_DIR) && \
 	find * -type d -exec zip -jr helm-$(VERSION)-{}.zip {} \; && \
 	cd -
```
krancour: The implementation of this target is all wrong.  It looks like these are the steps required for the missing `dist` target.

Makefile
```diff
@@ -1,47 +1,63 @@
...
 
-install: build
-	install -d ${DESTDIR}/usr/local/bin/
-	install -m 755 $(HELM_BIN) ${DESTDIR}/usr/local/bin/helm
+build-all:
+	${DEV_ENV_CMD} gox -verbose \
+	-ldflags "-X github.com/helm/helm/cli.version=${VERSION}" \
```
krancour: Can use the `$LDFLAGS` variable from up above (after its value is fixed).

Makefile
```diff
@@ -1,47 +1,63 @@
...
-	-os="linux darwin " \
-	-arch="amd64 386" \
-	-output="$(DIST_DIR)/{{.OS}}-{{.Arch}}/{{.Dir}}" .
 
-clean:
-	rm -rf $(DIST_DIR) $(BIN_DIR)
+# Builds the binary-- this should only be executed within the
+# containerized development environment.
 
-dist: build-all
```
krancour: This target should not be removed.  And it's implementation is currently under `build`, instead of here where it belongs.  When that is fixed, there's also an opportunity to containerize the steps involving `zip`.

Makefile
```diff
@@ -1,47 +1,63 @@
...
+binary-build:
+	make build-all
 	@cd $(DIST_DIR) && \
 	find * -type d -exec zip -jr helm-$(VERSION)-{}.zip {} \; && \
 	cd -
 
-install: build
```
krancour: This target should not be removed.  It's useful for installing a binary built from source onto the native system.  (i.e. This should remain non-containerized.)  Its pre-requisite should probably change to `native-build` to eliminate the possibility that the binary is built within the container for Linux/64 and then installed outside the container on Darwin/64 for it will not actually work.

Makefile
```diff
@@ -54,10 +70,10 @@ else
 endif
 
 quicktest:
-	$(PATH_WITH_HELM) go test -short ./ $(addprefix ./,$(GO_PACKAGES))
```
krancour: This `${PATH_WITH_HELM}` variable is very important.  The tests will not pass without it set.

Makefile
```diff
@@ -54,10 +70,10 @@ else
...
+	${DEV_ENV_CMD} go test -short ./ $(addprefix ./,$(GO_PACKAGES))
 
-test: test-style
-	$(PATH_WITH_HELM) go test -v ./ $(addprefix ./,$(GO_PACKAGES))
```
krancour: This ${PATH_WITH_HELM} variable is very important. The tests will not pass without it set.

Makefile
```diff
@@ -54,10 +70,10 @@ else
 endif
 
 quicktest:
-	$(PATH_WITH_HELM) go test -short ./ $(addprefix ./,$(GO_PACKAGES))
+	${DEV_ENV_CMD} go test -short ./ $(addprefix ./,$(GO_PACKAGES))
 
-test: test-style
```
krancour: The `test` target should still have `test-style` as a prerequisite.

Makefile
```diff
@@ -54,10 +70,10 @@ else
...
-	$(PATH_WITH_HELM) go test -v ./ $(addprefix ./,$(GO_PACKAGES))
+test:
+	${DEV_ENV_CMD} go test -v ./ $(addprefix ./,$(GO_PACKAGES))
 
 test-style:
```
krancour: This target has steps that can be containerized.  We shouldn't assume contributors have `go`, `gofmt`, and `golint` installed (and even if they do, we should not assume they have correct versions of each).

Makefile
```diff
@@ -1,47 +1,63 @@
...
-	install -d ${DESTDIR}/usr/local/bin/
-	install -m 755 $(HELM_BIN) ${DESTDIR}/usr/local/bin/helm
+build-all:
+	${DEV_ENV_CMD} gox -verbose \
+	-ldflags "-X github.com/helm/helm/cli.version=${VERSION}" \
+	-os="linux darwin " \
+	-arch="amd64 386" \
+	-output="$(DIST_DIR)/{{.OS}}-{{.Arch}}/{{.Dir}}" .
 
 prep-bintray-json:
```
krancour: There's another opportunity to containerize the usage of a less common tool here.  `jq` isn't so common that we can expect people have it installed-- I myself did not have it.  I have a PR open right now to add this to deis/docker-go-dev.

krancour:
I added lots and lots of line notes.  This is a more general comment, however...

When you containerize some of the more complex steps, you need to actually pass scripts to the container.  The pattern for that looks something like this:

```
${DEV_ENV_CMD} bash -c 'complex thing you want to do'
```

You will have to observe that pattern, for instance, for doing things like making modifications to the `$PATH` env var _within_ the container:

```
PATH_WITH_HELM = PATH=${DEV_ENV_WORK_DIR}/${BIN_DIR}:$$PATH
...
quicktest:
  ${DEV_ENV_CMD} bash -c '${PATH_WITH_HELM} go test -short ./ $(addprefix ./,${GO_PACKAGES})'
```

Note that the _single quotes_ are very, very important as they stop the local shell from interpreting `$PATH`.  You want that interpreted within the container.

krancour:
@chaitanyaenr for reference, many of the changes I mentioned in line notes have been applied in [this branch](https://github.com/krancour/helm/tree/containerize-the-build), but I haven't tested it too thoroughly.

Makefile
chaitanyaenr: @krancour Done

krancour:
Superseded by #445.  Closing.

PR #439, opened 2016-04-20
fix(fetch): copy the permisssion of the files during a helm fetch

kmala:


mboersma:
@kmala this seems like a smart change, but I did a before / after test and I see the same file permissions anyway. Is there a bug that this fixes?

kmala:
@mboersma  you can test using the https://github.com/deis/charts/pull/221 where there is script executing using the helm generate

arschles:
Is there a unit test to write/modify in here to ensure that file perms get copied?

mboersma:
> Is there a unit test...

There isn't anything that covers that package or `CopyDir()` yet.

util/helm_test.go
```diff
@@ -0,0 +1,42 @@
+package util
+
+import (
+	"testing"
```
mboersma: Could you `gofmt` this file?

PR #448, opened 2016-04-29
surface the env function

jhgaylor:
grabbing an env var tripped me up for a while when `ENV` was not a valid function. Would like to save others the hassle. Especially those not familiar with the go tools... like me ;)

### DIFF
--- a/docs/generate-and-template.md
+++ b/docs/generate-and-template.md
@@ -77,6 +77,7 @@ particularly useful ones are:
 - `b64enc <string>`: This base-64 encodes a string. Useful for secrets.
 - `randAlphaNum <int>`: This generates a random alphanumeric string of
   length INT
+- `env <string>`: Grabs the value of an environment variable. 
 
 But there are 40+ functions available. These functions are supplied by
 the built-in Go `text/template` package and the [Sprig template function


docs/generate-and-template.md
```diff
@@ -77,6 +77,7 @@ particularly useful ones are:
 - `b64enc <string>`: This base-64 encodes a string. Useful for secrets.
 - `randAlphaNum <int>`: This generates a random alphanumeric string of
   length INT
+- `env <string>`: Grabs the value of an environment variable.
```
krancour: "Grabs" sounds a little informal.  Could you please reword as "This retrieves the value of an environment variable."?

krancour:
Could you please also reword the commit to follow our usual conventions?  [CONTRIBUTING.md](https://github.com/helm/helm/blob/master/CONTRIBUTING.md) has a link out to [here](http://docs.deis.io/en/latest/contributing/standards/#commit-style-guide) outlining the commit style that helm has inherited from Deis.

Something along the following lines would be ideal: "docs(generate-and-template): Add mention of useful env function"

Thanks!

mboersma:
@jhgaylor could you update the Template Functions doc as suggested above, and amend the commit message? Something like:
```
docs(generate-and-template): mention the env function
```
Thanks again for helping with Helm Classic.

PR #458, opened 2016-05-03
doc(*): Update docs for helm-classic

### LINKED ISSUE #452: Update helm documentation for "helm classic"

krancour:
Fixes #452 

This reflects many changes that have not been implemented yet, but I'd prefer to not be submitting a mammoth PR.  If this means the docs at the head of the master branch are temporarily incorrect, I believe that is acceptable.  I have removed the readthedocs webhook from this repository that so that these changes will _not_ propagate to the _original_ (not to be conflated with "classic") helm docs at http://helm.readthedocs.io/en/latest/

The existing docs were also a bit stale, so some corrections were also made for accuracy.  For instance, the documented layout of the `helmc home` dir was all wrong.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -1,7 +1,15 @@
-# Helm Maintainers
+# Helm Classic Maintainers
 
-This document serves to describe the leadership structure of the Helm project, and to list the current
-project maintainers.
+|![](https://upload.wikimedia.org/wikipedia/commons/thumb/1/17/Warning.svg/156px-Warning.svg.png) | Helm Classic (the `helm/helm` repository) is **no longer actively developed** but will remain available until `kubernetes/helm` has stabilized.
+|---|---|
+
+Helm and [Deployment Manager](https://github.com/kubernetes/deployment-manager)
+have recently joined forces to make deploying and managing software on
+Kubernetes as simple as possible. The combined effort now lives in the Kubernetes GitHub organization at
+[kubernetes/helm](https://github.com/kubernetes/helm).
+
+This document serves to describe the leadership structure of the Helm Classic project, and to list the current
+project maintainers. _Maintainers of Helm Classic currently perform bug fixes and critical maintenance only._
 
 # What is a maintainer?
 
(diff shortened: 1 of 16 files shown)


krancour:
cc @kmala

docs/generate-and-template.md
```diff
@@ -173,10 +173,11 @@ However, environment variables are expanded.
 Along with any existing environment variables, the following variables
 are specially defined:
 
-- HELM_HOME: The Helm home directory
-- HELM_DEFAULT_REPO: The repository alias for the default repository.
-- HELM_GENERATE_FILE: The present file's name
-- HELM_GENERATE_DIR: The absolute path to the chart directory of the present chart
+- `HELM_HOME`: The Helm Classic home directory
+    - Although the value of the `HELMC_HOME` environment variable, if defined, may play a role in defining the specially expanded `HELM_HOME` variable, the two are not equivalent. To produce charts that remain compatible with the _original_ Helm tool, which has now become Helm Classic, only the `HELM_HOME` variable should be used within `helm:generate` headers.
```
mboersma: I couldn't quite understand this part. I can't tell if or how `$HELMC_HOME` is used, and if `$HELM_HOME` is preferred for backward-compatibility, do we need to mention `$HELMC_HOME` at all?

mboersma:
Good work here. I had one comment but LGTM either way.

docs/generate-and-template.md
krancour: @mboersma Ya.  It's confusing.  :(  The helmc binary itself is (or rather, _will be_) sensitive to the `HELMC_HOME` environment variable, which can _optionally_ be used to specify a custom home (rather than the default).

_However..._ to be backwards compatible with generator directives that appear in _existing_ charts (created for the _original_ helm), `HELM_HOME` (no `C`) will continue to be expanded there as it always has.

If someone opted to use `HELMC_HOME` in a generator directive, that chart wouldn't be usable by people using an older version of helm because the older version does not expand that variable.

I know that's all confusing.  fwiw, the _new_ (kubernetes/)helm wants to use `HELM_HOME` defaulted to `~/.helm`, which is why that team had +1'ed helm classic having its own distinct home-- so both tools can co-exist for some period of time.

I'm not sure how to make all of this clear in the docs without being as verbose as I just was.  :(  I am open to all suggestions on rewording.

MAINTAINERS.md
```diff
@@ -1,7 +1,15 @@
...
+
+Helm and [Deployment Manager](https://github.com/kubernetes/deployment-manager)
+have recently joined forces to make deploying and managing software on
+Kubernetes as simple as possible. The combined effort now lives in the Kubernetes GitHub organization at
+[kubernetes/helm](https://github.com/kubernetes/helm).
+
+This document serves to describe the leadership structure of the Helm Classic project, and to list the current
+project maintainers. _Maintainers of Helm Classic currently perform bug fixes and critical maintenance only._
 
 # What is a maintainer?
```
kmala: nitpick...shouldn't this be `who is a maintainer?`

krancour: I think "what" is appropriate here since what follows is a definition of the role rather than a list of who currently fulfills the role.

docs/generate-and-template.md
krancour: I will continue to think about how this can be clarified further, but I am going to merge so this effort doesn't fall behind.

PR #460, opened 2016-05-03
ref(*): Use HELMC_HOME instead of HELM_HOME

### LINKED ISSUE #459: Update HELM_HOME -> HELMC_HOME for helm-classic
Update `HELM_HOME` env var to `HELMC_HOME` and its default value from `~/.helm` to `~/.helmc`.

krancour:
Fixes #459 

cc @kmala

### DIFF
--- a/action/generate.go
+++ b/action/generate.go
@@ -20,6 +20,9 @@ func Generate(chart, homedir string, exclude []string, force bool) {
 	}
 	chartPath := util.WorkspaceChartDirectory(homedir, chart)
 
+	// Although helmc itself may use the new HELMC_HOME environment variable to optionally define its
+	// home directory, to maintain compatibility with charts created for the ORIGINAL helm, we
+	// continue to support expansion of these "legacy" environment variables, including HELM_HOME.
 	os.Setenv("HELM_HOME", homedir)
 	os.Setenv("HELM_DEFAULT_REPO", mustConfig(homedir).Repos.Default)
 	os.Setenv("HELM_FORCE_FLAG", strconv.FormatBool(force))
--- a/action/plugin.go
+++ b/action/plugin.go
@@ -24,6 +24,10 @@ func Plugin(homedir, cmd string, args []string) {
 	if abs, err := filepath.Abs(homedir); err == nil {
 		homedir = abs
 	}
+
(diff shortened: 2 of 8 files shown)


action/generate.go
```diff
@@ -20,6 +20,9 @@ func Generate(chart, homedir string, exclude []string, force bool) {
 	}
 	chartPath := util.WorkspaceChartDirectory(homedir, chart)
 
+	// Although helmc itself may use the new HELMC_HOME environment variable to optionally define its
+	// home directory, to maintain compatibility with charts created for the ORIGINAL helm, we
+	// continue to support expansion of these "legacy" environment variables, including HELM_HOME.
 	os.Setenv("HELM_HOME", homedir)
 	os.Setenv("HELM_DEFAULT_REPO", mustConfig(homedir).Repos.Default)
 	os.Setenv("HELM_FORCE_FLAG", strconv.FormatBool(force))
```
kmala: shouldn't we be setting the values for the  HELMC_HOME and others so that when new charts are created and uses $HELMC_HOME it works?

krancour: New charts should continue to use `HELM_HOME`.  If people start using `HELMC_HOME` in charts, those chars won't work with older / original helm.  This is addressed in the updated documentation I have also PR'ed.  It ensures both backwards and forwards compatibility.

Re: whether we might want to _additionally_ set `HELMC_HOME`, I believe the answer is no.  Existing environment variables _do_ get expanded here.  This means if someone has `HELMC_HOME` set to a custom value, it will be properly inherited and interpreted by any possible invocation of the `helmc` binary in the generator header.  If it is absent, the invocation of the `helmc` binary will use the default helmc home dir, just as the initial invocation of `helmc` (by the user) would have.

The only thing we'd gain by setting `HELMC_HOME` is that chart authors would be able to use it for something like the following:

```
#helm:generate sed -i s/foo/bar/ $HELMC_HOME/blah/blah
```

And that is precisely the thing we _don't_ want chart authors doing because it will break their chart's compatibility with older / original helm.  I feel that if we don't want them doing that, we shouldn't encourage that mistake by unnecessarily injecting a value into the `HELMC_HOME` env var when generators are invoked.

Issue #450, opened 2016-04-29
Meta: Transition helm to "helm classic"


krancour:
This is a meta issue to track other discreet tasks that need to be completed to:

* Change this repository's name to "helm-classic".
* Stop building a `helm` binary from this repository and _start_ building a `helmc` binary instead.
    * This is a tactical move to create a more obvious distinction between this repository (and the tool built from it) and the [kubernetes/helm](https://github.com/kubernetes/helm) repository (and the next generation `helm` tool built from that).  This is intended to start addressing questions like that posed in #427.

Discrete tasks:

Helm tasks - pre-release:

- [x] Change the name of this repository to "helm-classic".  ~~(This should probably come last.)~~ #449 
    - This can happen at any time since we're using a containerized build.  IMO, it makes the most sense to   do it after we have merged updated docs, package names, home, and binary.
- [x] Rename packages. #451 
- [x] Update helm documentation. #452
- [x] Update build to produce `helmc` binary. #453
    - Will likely require changes in bintray as well.
    - Ensure generators in existing charts that recursively invoke the `helm` binary, somehow invoke `helmc` instead _without_ requiring modifications to charts.  i.e. Existing charts must not break when a user switches to `helmc`.
- [x] ~~_Maybe_~~ update `HELM_HOME` env var to `HELMC_HOME` and its default value from `~/.helm` to `~/.helmc`.  This is largely dependent on how safely `helmc` and the new `helm` can or cannot share a common home.  ~~Ping @technosophos.  Thoughts?~~ Per @michelleN, we _will_ do this. #459 
- [x] Update interactive help for `helmc`. #469 
- [x] Add migration docs-- explain how to get from "original" Helm to Helm Classic and how Helm Classic can co-exist with kubernetes/helm for as long as needed going forward.
    - This works because they use different home dirs.
    - Migration from original Helm should address moving / renaming existing home dir to clear the way for the new Helm.  (e.g. `mv $(helm home) $(helmc home) && rm $(which helm)`) #463
- [x] Release Helm Classic 0.7.0

Helm tasks - post-release:

- [x] Update Helm Classic change log
- [x] Update / deploy install script. https://github.com/engineyard/get-helm/pull/1
- [x] Update https://helm.sh. https://github.com/helm/helm-www/pull/21

Other related tasks:

- [x] Update [deis/workflow](https://github.com/deis/workflow) documentation to direct users to install via `helmc`. https://github.com/deis/workflow/issues/211
[...]

PR #467, opened 2016-05-05
ref(helmc): Rename the uploaded binary files to start with helmc

kmala:
ref #450 

cc @krancour

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ clean:
 	rm -rf ${DIST_DIR} ${BIN_DIR}
 
 dist: build-all
-	${DEV_ENV_CMD} bash -c 'cd ${DIST_DIR} && find * -type d -exec zip -jr helm-${VERSION}-{}.zip {} \;'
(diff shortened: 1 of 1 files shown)


krancour:
Do we know how this script gets out to https://get.helm.sh?

krancour:
Per @sgoings this is not the actual source of the install script found at https://get.helm.io.

Should we consider a new PR that removes the script from here to avoid this same confusion in the future?

sgoings:
I'd be a fan of removing the install script from this repository or getting auto deployment logic from the internal repo into this one. 😄

krancour:
I'm going to close this in favor of a forthcoming PR that removes the script.

kmala:
This PR also has a small fix which i think is needed to name the binary zip file to helmc unless we don't wanna do it?

krancour:
We should.  Maybe reopen or resubmit with just that one change?

kmala:
@krancour kept just that one change and renamed the PR

PR #468, opened 2016-05-05
chore(*): update go to 1.6.2 and glide to 0.10.2

mboersma:
`make bootstrap` kept failing for me, so I took out the big hammer and updated `go` and `glide` and regenerated the dependency lock file. Works For Me&trade; now, although honestly the original errors may have been transient and not specifically fixed by this. 😄


krancour:
LGTM... even if not completely necessary, it's good to keep dependencies up to day.  I especially wanted the newer go-dev as well because I am hoping to use its changelog generator.

arschles:
:+1: to what @krancour said

:shipit:

PR #470, opened 2016-05-05
chore(*): Update comments and help strings for Helm Classic

### LINKED ISSUE #469: Update interactive help for helmc
Interactive help and output messages should be updated to reflect the product's and binary's new names.

krancour:
Fixes #469

This PR updates help, output messages, comments, and a few other errant, outdated occurrences of "helm" or "Helm" with "helmc" and "Helm Classic" wherever appropriate.

### DIFF
--- a/cli/helm.go
+++ b/cli/helm.go
@@ -15,21 +15,21 @@ import (
 // and should not be trusted.
 var version = "0.1.0"
 
-const globalUsage = `The Kubernetes package manager
+const globalUsage = `Helm Classic - A Kubernetes package manager
 
-To begin working with Helm, run the 'helm update' command:
+To begin working with Helm Classic, run the 'helmc update' command:
 
-$ helm update
+$ helmc update
 
 This will download all of the necessary data. Common actions from this point
 include:
 
-- helm help COMMAND: see help for a specific command
-- helm search: search for charts
-- helm fetch: make a local working copy of a chart
-- helm install: upload the chart to Kubernetes
+- helmc help COMMAND: see help for a specific command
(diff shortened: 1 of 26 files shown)


cli/helm.go
```diff
@@ -15,32 +15,32 @@ import (
 // and should not be trusted.
 var version = "0.1.0"
 
-const globalUsage = `The Kubernetes package manager
+const globalUsage = `Helm Classic - The original Kubernetes package manager
```
arschles: "The original Kubernetes package manager" sounds a little weird here, like it's claiming to be the first (i.e. 1000s of chains all claiming to be "The original pancake house"). Just a thought - not really a big deal IMO

krancour: I was only trying to match it to the documentation.  We've used that tagline in a few places to try to distinguish Helm Classic from the new Helm, since this is something that as confused the community a bit up until now.  If we change this here, we should probably change it in other places as well.  Let's run this by @slack before I merge.

slack: You could just keep it as `The Kubernetes Package Manager` or just `A Kubernetes Package Manager`.

krancour: I'll drop original and will go with "a" over "the".  The indefinite article more appropriately acknowledges the existence of _other_ Kubernetes package managers. I wouldn't say that's terribly important, except that the _new_ Helm is one of them!

krancour: I'm going to make the same change to the docs as well-- keeping that in _this_ PR so that this one PR enforced consistency in the tagline.

PR #474, opened 2016-05-10
feat(manifest|action): handle "helm-keep" annotation

### LINKED ISSUE #465: Tweak `helm [un]install` to make upgrades easier
To upgrade Deis Workflow (as one example of a complex Chart), it's required that cluster ingress is preserved. Essentialy any DNS pointing toward a `LoadBalancer` should not need to be recreated. That requires that the `Namespace` in which it lives is also not destroyed.

This is possible by running a passel o' `kubectl` commands (here is [a quick-and-dirty version](https://gist.github.com/mboersma/10a7dece39900c8760ecd9c595032afe)). I propose we modify `helmc` behavior slightly to accomodate this.

*Edit*: this proposal changed quite a bit from here--see the end of this discussion and #474.

`helmc uninstall --upgrading` would change the current `uninstall` behavior to leave intact:

1. k8s `Service`s containing either a `LoadBalancer` or `NodePort`
2. `Namespace`s

`helmc install --upgrading` would change the current `install` behavior to use `kubectl apply` instead of `kubectl create`. This should create any missing resources while updating the definitions of those that already exist.

`kubectl apply` may not effect runtime changes on an existing `Service`, but my assumption is that will be rare and can be worked around. The onus is on on the Chart developer to maintain compatibility with `--upgrade`: essentially "don't make important changes to an externally routed `Service` definition without providing additional upgrade instructions."

While this is designed to solve the immediate needs of upgrading Deis Workflow charts, it's a general mechanism. Here's a typical upgrade process based on this change:

```console
$ helmc uninstall --upgrading -n deis -y workflow-beta3
$ helmc fetch deis/workflow-beta4
$ helmc generate -x manifests workflow-beta4
$ helmc install --upgrading workflow-beta4
```

ping @smothiki @krancour 

See also #324.

mboersma:
This modifies `helmc` to treat manifests with a `helm-keep: "true"` annotation specially:
- On `install`, they are treated with `kubectl apply` rather than `kubectl create`
- On `uninstall`, they are skipped with a warning

This behavior should allow crucial manifests to persist over a `helmc uninstall <version_one> && helmc install <version_two>` upgrade scenario. In particular, an externally routed service can be annotated with `helm-keep` so that DNS does not have to be recreated for its `LoadBalancer`.

Closes #465.

### DIFF
--- a/action/uninstall.go
+++ b/action/uninstall.go
@@ -127,6 +127,14 @@ func uninstallKind(kind []*manifest.Manifest, ns, ktype string, dry bool, client
 		if dry {
 			log.Msg("%s/%s", ktype, o.Name)
 		} else {
+			// If it's a keeper manifest, skip uninstall.
+			if data, err := o.VersionedObject.JSON(); err == nil {
+				if manifest.IsKeeper(data) {
+					log.Warn("Not uninstalling %s %s because of \"helm-keep\" annotation.\n"+
+						"---> Use kubectl to uninstall keeper manifests.\n", ktype, o.Name)
+					continue
+				}
+			}
 			out, err := client.Delete(o.Name, ktype, ns)
 			if err != nil {
 				log.Warn("Could not delete %s %s (Skipping): %s", ktype, o.Name, err)
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 
 	"github.com/helm/helm-classic/codec"
 	"github.com/helm/helm-classic/log"
@@ -123,4 +124,11 @@ func ParseDir(chartDir string) ([]*Manifest, error) {
 	}
 
 	return files, filepath.Walk(dir, walker)
+}
+
+// IsKeeper returns true if a manifest has a "helm-keep": "true" annotation.
+func IsKeeper(data []byte) bool {
+	// Look for ("helm-keep": "true") up to 10 lines after ("annotations:").
+	var keep = regexp.MustCompile(`\"annotations\":\s+\{(\n.*){0,10}\"helm-keep\":\s+\"true\"`)
+	return keep.Match(data)
 }
--- a/action/action_test.go
+++ b/action/action_test.go
@@ -9,6 +9,10 @@ func (r TestRunner) ClusterInfo() ([]byte, error) {
 	return r.out, r.err
 }
 
+func (r TestRunner) Apply(stdin []byte, ns string) ([]byte, error) {
+	return r.out, r.err
+}
+
 func (r TestRunner) Create(stdin []byte, ns string) ([]byte, error) {
 	return r.out, r.err
 }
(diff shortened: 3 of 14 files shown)


action/uninstall.go
```diff
@@ -127,6 +127,14 @@ func uninstallKind(kind []*manifest.Manifest, ns, ktype string, dry bool, client
 		if dry {
 			log.Msg("%s/%s", ktype, o.Name)
 		} else {
+			// If it's a keeper manifest, skip uninstall.
+			if data, err := o.VersionedObject.JSON(); err == nil {
+				if manifest.IsKeeper(data) {
+					log.Warn(`Not uninstalling %s %s because of "helmc: keep" annotation.
```
krancour: I think this message need to change to match the annotation you actually used: `helm-keep: true`.

manifest/manifest.go
```diff
@@ -124,3 +125,9 @@ func ParseDir(chartDir string) ([]*Manifest, error) {
 
 	return files, filepath.Walk(dir, walker)
 }
+
+// IsKeeper returns true if a "helm-keep: true" annotation is present in manifest bytes.
+func IsKeeper(data []byte) bool {
+	var keep = regexp.MustCompile(`\"annotations\": \{(\n.*){0,10}\"helm-keep\": true`)
```
krancour: Won't this be true even if it's commented?  That could be a problem.

mboersma: You mean if there's a commented-out line like `# "helm-keep": true`? Since we are examining only the reconstituted JSON bytes of the manifest in this case, I don't think that will occur.

But it is a sloppy regex, mostly intentionally so. It says "if you see `"helm-keep": true` in the ten lines following `"annotations":`, assume it's a keeper. I didn't want to get into walking the actual structured data, so searching the byte stream seemed like a reasonable approach.I would love suggestions for how to tighten up the regex.

krancour: Oh... nevermind... this is JSON, isn't it?  No comments.

mboersma: Maybe this would be a reason to revisit the idea of putting this in the file header instead of as an annotation? That didn't look quite as straightforward to implement, but could be safer to parse out.

krancour: I agree it's not as straightforward to implement if it's in a header instead.  Ease of implementation set aside, I'm torn on which is the "right" way to do this.  I don't think I have enough of an opinion on it yet.

mboersma:
Updated docs with a section describing "keeper" manifests in awesome.md.

jchauncey:
Like this change

bacongobbler:
we'll have to make sure to `helm-keep` our namespace as well as the router and whatever else we want to keep. LGTM!

krancour:
Have we tested this out against Workflow charts that are modified as such?

mboersma:
@krancour I have tested this with deis/charts#264, but I would certainly welcome more testing if anyone has time.

arschles:
are there docs in deis/workflow that explain how to upgrade?

mboersma:
> are there docs in deis/workflow that explain how to upgrade?

@arschles not yet, but I've added the "needs docs" tag to deis/charts#264 so it doesn't get missed. (I need to figure out how that PR can be merged without disrupting CI as well, since `helmc uninstall` currently expects that it will remove everything.)

PR #476, opened 2016-05-23
fix(glide.yaml): point to mirror of dec/inf package

mboersma:


glide.yaml
```diff
@@ -15,6 +15,8 @@ import:
   - pkg
 - package: speter.net/go/exp/math/dec/inf
   version: 42ca6cd68aa922bc3f32f1e056e61b65945d9ad7
```
arschles: @mboersma does this version still exist in the aliased repo?

mboersma: It seems to, since `helmc` builds.... And [yes, it exists](https://github.com/belua/inf/commit/42ca6cd68aa922bc3f32f1e056e61b65945d9ad7).

